PR #171, opened 2022-05-26
CLI refactor

danecando:
Updates the CLI to get the list of starter kits from a raw JSON file hosted on GitHub `starter-kits.json`. Currently entries will need to be added manually. Also this won't really work until the repo is public due to short lived tokens for raw files. It will probably be ideal to write a script that generates the JSON file and put it in a GH action to automate it


danecando:
Reminder to self: publish to npm after merge :)

PR #184, opened 2022-06-14
Vue3 apollo quasar/state management with apollo

hdJerry:
- Added state management with apollo
- Test broke so i had to fix it


starters/vue3-apollo-quasar/src/components/NumberCounter/NumberCounter.vue
```diff
@@ -36,15 +36,42 @@
...
+import { computed, defineComponent } from 'vue';
 
 export default defineComponent({
   name: 'NumberCounter',
 });
 </script>
 
 <script lang="ts" setup>
-import { useCounterStore } from 'src/stores';
+import { counts } from 'src/variables/counts';
```
AllanJeremy, line 47: I think this is an unnecessary abstraction, we can move this logic into the number counter component. Or, if this is meant to serve the same purpose as a counter store, then we should be able have it within this component (in which case it isn't global)

starters/vue3-apollo-quasar/src/components/NumberCounter/NumberCounter.vue
```diff
@@ -36,15 +36,42 @@
...
+  query Counter {
+    count @client
+  }
+`;
+
+const { result, loading } = useQuery(COUNTERS_QUERY);
+
+const count = computed(() => (result.value ? result.value.count : 0));
+
+const updateCount = (countValue: number) => {
```
AllanJeremy, line 61: This is counter-intuitive.
Rather than testing if `updateCount` does addition, subtraction and reset, create individual functions for `increment` `decrement` and `reset`

Also, the use of switch in this case would be unnecessary. Needs refactoring.

starters/vue3-apollo-quasar/src/variables/counts.js
```diff
@@ -0,0 +1,3 @@
+import { makeVar } from '@apollo/client';
+
+export const counts = makeVar(0);
```
AllanJeremy, line 3: If the way forward is to keep this as an abstraction, there are 2 main suggestions:
- Rename the `variables` directory to something like `globals` or `store` to indicate that it is a directory that holds global state.

- Move mutation functionality into this directory. No point in decoupling it from this and coupling it to the `NumberCounter` component. _View more comments on the same in the `NumberCounter` component review._

hdJerry: changing variables to store doesn't look good as it defeats what the Reactive variable is.

starters/vue3-apollo-quasar/src/components/NumberCounter/NumberCounter.vue
hdJerry: I think the abstraction is okay.

hdJerry:
what i could do is to take the increment, decrement and reset to the place this reactive variable file, so i just export the functions

starters/vue3-apollo-quasar/src/components/NumberCounter/NumberCounter.vue
AllanJeremy: The way I see it, we have 2 options:

- The abstraction is unnecessary if the mutation logic is coupled to the `NumberCounter` component. In this case move the logic into the `NumberCounter` component. This negates the point of using apollo for global state though. No need to do 50-50 with the implementation

- The mutation logic gets moved into the abstraction and out of `NumberCounter` to make it truly global functionality.

starters/vue3-apollo-quasar/src/variables/counts.js
AllanJeremy: `store` was an example. What I'm trying to illustrate is that we need a name that is more self-descriptive. 
What does `variables` mean? Is this a convention? Since this will be used by many people, we do not want to introduce new conventions that are not necessarily known by the general community.

The suggestion of `store` or `globals` carries forward user pre-existing knowledge

hdJerry: Renamed to globals

starters/vue3-apollo-quasar/src/components/NumberCounter/NumberCounter.vue
hdJerry: I think that's what i have done. thansk

starters/vue3-apollo-quasar/src/variables/counts.js
hdJerry: variables is kinda a convention, but globals is fine

AllanJeremy (approved):
LGTM 🚀

ihardz (approved):

starters/vue3-apollo-quasar/src/components/NumberCounter/NumberCounter.vue
hdJerry: Copy that

PR #221, opened 2022-06-20
[Website] Fix Production Build by Changing MobileNavigation Import Path

### LINKED ISSUE #220: [angular-ngrx-scss] Kit not deployed to npm correctly

Reshurum:
Resolves #220 

The build for whatever reason expects a '.tsx' for the import path to MobileNavigation. The linter seems upset now after making this fix, but the build now functions as expected.

The change that introduced this regression can be found [here](https://github.com/thisdot/starter.dev/commit/ce393f68efcaa6e2c82085c7f8c2cc973d70397c#diff-164c9e7ead49ac55ce388ddb4647642b45501848d28538006738705f1c0eb81fL4-R4). This regression was not introduced by the merge of the new Ng-NgRx-SCSS starter as we previously thought. Also here are [the build logs](https://us-east-1.console.aws.amazon.com/amplify/home?region=us-east-1#/d2zmkukv1jdnvq/main/48) associated with the aforementioned change. It appears that this got introduced a while back but it got missed.

### DIFF
--- a/packages/website/src/pages/frameworks/index.astro
+++ b/packages/website/src/pages/frameworks/index.astro
@@ -1,7 +1,7 @@
 ---
 import BaseLayout from '../../layouts/BaseLayout.astro';
 import TechItem from '../../components/TechItem.astro';
-import MobileNavigation from '../../components/MobileNavigation';
+import MobileNavigation from '../../components/MobileNavigation.tsx';
 import { TECHNOLOGIES } from '../../config';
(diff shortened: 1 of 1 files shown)


danecando (approved):
That's annoying but if it needs to be changed to work then what are you gonna do

In that linked commit I had changed the tsconfig to one that they have on their website. I've switched these from no extension to extension several times. It seems like sometimes astro wants it and sometimes it doesn't so I dono

PR #229, opened 2022-06-20
[Ng-NgRx-SCSS] Pin all Dependency Versions

### LINKED ISSUE #223: [angular-ngrx-scss] Lock dependency versions

Reshurum:
Resolves #223 

This change also changes the global `.gitignore` as specified in the issue, "Add to starter.dev gitignore but not local gitignore". This change should work for all of the starters for npm, yarn and pnpm.


Reshurum:
Question: There was no lock file in the starter itself because we are using yarn workspaces. I assume the requirement to remove the existing lockfile is for the lockfiles in the starters themselves and not the global one also used with the website package?

Issue #195, opened 2022-06-17
[angular-apollo-tailwind] Correct linting errors in generated projects


dustinsgoodman:
# Background

When I start a new project, I want 0 linting errors to start. There are several minor linting errors in the default scaffold.

![Image](https://user-images.githubusercontent.com/1815379/174395911-65a971a9-077f-4480-b10e-bd0b965953ed.png)

```
yarn lint
yarn run v1.22.18
$ npx eslint "src/**/*.{js,jsx,ts,tsx,html,css,scss}" --quiet --fix

/Users/dustingoodman/development/starter-testing/ng-apollo-test/src/app/home/home.component.ts
  11:5  error  Lifecycle methods should not be empty  @angular-eslint/no-empty-lifecycle-method

/Users/dustingoodman/development/starter-testing/ng-apollo-test/src/stories/button.component.ts
  6:8  error  Parsing error: '>' expected

/Users/dustingoodman/development/starter-testing/ng-apollo-test/src/stories/button.css
  1:1  error  Parsing error: Unexpected token .

/Users/dustingoodman/development/starter-testing/ng-apollo-test/src/stories/header.component.ts
   5:15  error  The selector should start with one of these prefixes: "app" (https://angular.io/guide/styleguide#style-02-07)                                                                                                                       @angular-eslint/component-selector
  69:5   error  Output bindings, including aliases, should not be named "on", nor prefixed with it (https://angular.io/guide/styleguide#style-05-16)                                                                                                @angular-eslint/no-output-on-prefix
  72:5   error  Output bindings, including aliases, should not be named "on", nor prefixed with it (https://angular.io/guide/styleguide#style-05-16)                                                                                                @angular-eslint/no-output-on-prefix
  75:5   error  Output bindings, including aliases, should not be named "on", nor prefixed with it (https://angular.io/guide/styleguide#style-05-16)                                                                                                @angular-eslint/no-output-on-prefix
  47:27  error  Parsing error: Unexpected closing tag "storybook-button". It may happen when the tag has already been closed by another tag. For more info see https://www.w3.org/TR/html5/syntax.html#closing-elements-that-have-implied-end-tags  prettier/prettier

/Users/dustingoodman/development/starter-testing/ng-apollo-test/src/stories/header.css
  1:1  error  Parsing error: Unexpected token .

/Users/dustingoodman/development/starter-testing/ng-apollo-test/src/stories/page.component.ts
[...]

PR #232, opened 2022-06-21
angular-apollo-tailwind. Correct linting errors in generated projects

ihardz:
[#195 angular-apollo-tailwind. Correct linting errors in generated projects](https://github.com/thisdot/starter.dev/issues/195)


starters/angular-apollo-tailwind/src/styles.scss
```diff
@@ -1,5 +1,5 @@
 /* You can add global styles to this file, and also import other style files */
 
-@tailwind base;
-@tailwind components;
-@tailwind utilities;
+@import "tailwindcss/base";
+@import "tailwindcss/components";
+@import "tailwindcss/utilities";
```
dustinsgoodman, line 5: I don't think this is the right fix here. We may need to install the tailwind plugin.

dustinsgoodman (changes requested):

dustinsgoodman:
@ihardz Also - I think I'd like stylelint added as part of this. I think not having CSS linting is problematic so just removing it via eslint isn't right IMO

starters/angular-apollo-tailwind/src/styles.scss
ihardz: Corrected.

dustinsgoodman (approved):

PR #237, opened 2022-06-22
[angular-apollo-tailwind] Remove default stories #196

### LINKED ISSUE #196: [angular-apollo-tailwind] Remove default stories
# Background
Storybook ships with a few default stories by default. We do not want to include these stories as part of our kits.

Image

# Acceptance
Remove default stories and related files.

ihardz:
Resolves #196 

> **Background**
>
> Storybook ships with a few default stories by default. We do not want to include these stories as part of our kits.
> 
> **Acceptance**
> Remove default stories and related files.


starters/angular-apollo-tailwind/.gitignore
```diff
@@ -40,3 +40,7 @@ testem.log
 # System files
 .DS_Store
 Thumbs.db
+
+# Auto-Generated
+
+documentation.json
```
dustinsgoodman, line 46: Do we still need to be generating this? It's in the package.json scripts but I wonder if we really need it.

dustinsgoodman (approved):
I'm good with this PR as is but left a question about the documentation.json

starters/angular-apollo-tailwind/.gitignore
dustinsgoodman: I guess I don't understand why this is in the kits. Does anyone know? @morgnism @BrettZeidler

BrettZeidler: yeah not sure why specifically -- I think it was added before I was brought into the project

would be best to .gitignore it though I'd say

Issue #209, opened 2022-06-17
[angular-ngrx-scss] update loader styles on fetch example


dustinsgoodman:
# Background
I want to standardize all the loading components across all kits.

# Acceptance
While the component is loading, show a loading state as a skeleton.

# Note
https://dev.to/jobpick/how-to-create-a-skeleton-loader-in-tailwindcss-38gh

PR #250, opened 2022-06-24
[angular-ngrx-scss] - Update loader styles on fetch example

SDaian:
Closes: #209

### DIFF
--- a/starters/angular-ngrx-scss/src/app/fetch-example/fetch-example.component.html
+++ b/starters/angular-ngrx-scss/src/app/fetch-example/fetch-example.component.html
@@ -7,7 +7,7 @@
   </ng-container>
 
   <ng-template #loadingTemplate>
-    <p class="fetch-example__greeting">Loading message...</p>
+    <app-loader></app-loader>
   </ng-template>
 
   <nav>
--- a/starters/angular-ngrx-scss/src/app/fetch-example/loader/loader.component.html
+++ b/starters/angular-ngrx-scss/src/app/fetch-example/loader/loader.component.html
@@ -1,0 +1,1 @@
+<div class="loader"></div>
--- a/starters/angular-ngrx-scss/src/app/fetch-example/loader/loader.component.scss
+++ b/starters/angular-ngrx-scss/src/app/fetch-example/loader/loader.component.scss
@@ -1,0 +1,24 @@
+:host {
+  display: block;
+}
+.loader {
+  &:before {
+    content: '\00a0 ';
+  }
+  border-radius: 0.375rem;
+  border-width: 2px;
+  font-size: 1.2rem;
+  margin: 1em 0;
+
+  background-color: #d1d5db;
+  animation: pulse 2s cubic-bezier(0.4, 0, 0.6, 1) infinite;
+  @keyframes pulse {
+    0%,
+    100% {
+      opacity: 1;
+    }
+    50% {
+      opacity: 0.5;
+    }
+  }
+}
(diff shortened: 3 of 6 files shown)


lindakatcodes (approved):
I think this looks good to me! Works well and looks similar to the example in the linked post. 👍

starters/angular-ngrx-scss/src/app/fetch-example/loader/loader.component.html
```diff
@@ -0,0 +1,8 @@
+<div class="loading">
+  <div class="loading__section">
+    <div class="loading__wrapper">
+      <div class="loading__item" [style.height]="height" [style.width]="width">
+      </div>
+    </div>
+  </div>
+</div>
```
Reshurum, line 8: This DOM structure can be simplified greatly with the stylesheet changes I mention elsewhere in this review.

```suggestion
<div class="loader"></div>
```

starters/angular-ngrx-scss/src/app/fetch-example/loader/loader.component.scss
```diff
@@ -0,0 +1,30 @@
...
+  &__wrapper {
+    display: flex;
+    margin: 0.375rem 0;
+    flex-direction: column;
+  }
+  &__item {
+    background-color: #D1D5DB;
+    border-radius: 0.375rem;
+  }
+}
```
Reshurum, line 30: Using flexbox for horizontal centering of this skeleton is overkill. This can be done with a block element and an automatic margin.

Also the height is set by component code at the moment which is not ideal. This can be changed to use the same font size of the text we're loading with a `:before` selector as far as the height is concerned. In regards to the width, we can use the height / max-height CSS properties set on the parent element itself.

```suggestion
:host {
  display: block;
}

.loader {
  &:before {
    content: '\00a0 ';
  }

  border-radius: 0.375rem;
  border-width: 2px;
  font-size: 1.2rem;
  margin: 1em 0;
  
  background-color: #d1d5db;
  animation: pulse 2s cubic-bezier(0.4, 0, 0.6, 1) infinite;

  @keyframes pulse {
    0%,
    100% {
      opacity: 1;
    }
    50% {
      opacity: 0.5;
    }
  }
}
```

starters/angular-ngrx-scss/src/app/fetch-example/loader/loader.component.ts
```diff
@@ -0,0 +1,11 @@
+import { Component, Input } from '@angular/core';
+
+@Component({
+  selector: 'app-loader',
+  templateUrl: './loader.component.html',
+  styleUrls: ['./loader.component.scss']
+})
+export class LoaderComponent {
+  @Input() width = '140px';
+  @Input() height = '24px'
```
Reshurum, line 10: If my CSS suggestions are made, then these inputs are no longer needed.

starters/angular-ngrx-scss/src/app/fetch-example/fetch-example.component.html
```diff
@@ -7,7 +7,7 @@ <h1 class="fetch-example__header">NgRx Fetch Data from API</h1>
   </ng-container>
 
   <ng-template #loadingTemplate>
-    <p class="fetch-example__greeting">Loading message...</p>
+    <app-loader [width]="'260px'"></app-loader>
```
Reshurum, line 10: We should avoid setting the width of an element using component properties if possible.

Reshurum (changes requested):
If my suggestions are made, the width of the loader can be set in the fetch example page using the following CSS instead of using component properties:
```scss
.fetch-example {
  app-loader {
    max-width: 260px;
    margin: 0 auto;
  }
}
```

Also please format any files that you add / change with prettier before submission.

PR #256, opened 2022-06-28
[angular-apollo-tailwind] project specific naming variables need overrides

### LINKED ISSUE #193: [angular-apollo-tailwind] project specific naming variables need overrides
angular.json and other key project files do not replace their path with the actual project name. Needs to replace these values so scripts work.

ihardz:
Resolves #193

### DIFF
--- a/packages/create-starter/src/index.ts
+++ b/packages/create-starter/src/index.ts
@@ -49,7 +49,23 @@ export async function main() {
   }
 
   const repoPath = `thisdot/starter.dev/starters/${options.kit}`;
-  const destPath = path.join(process.cwd(), options.name);
+  let parentDir = process.env.INIT_CWD || process.cwd();
+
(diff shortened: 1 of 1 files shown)


danecando (changes requested):
I don't think any of this code is needed. The destPath is just relative to wherever they run the command from. The ticket is about replacing values in angular.json that use the kit/directory name to whatever the user chose for the project name (options.name)

So same thing that is already in there for the package.json file but for replacing name values in angular.json

danecando:
I would discard what you have now and just add whatever is needed to replace the name values for items in the angular.json file like we do currently for the package.json file

dustinsgoodman:
@danecando has the right sense of the problem. It's just that the angular.json file isn't inheriting the project name and uses the default provided by the repo which is wrong.

PR #267, opened 2022-07-08
[next-react-query-tailwind] Lock dependency versions

### LINKED ISSUE #225: [next-react-query-tailwind] Lock dependency versions
# Background
See conversation on https://github.com/thisdot/starter.dev/issues/200

# Acceptance
- Pin all dependency versions in the package.json
- Remove lock file
- Add to starter.dev gitignore but not local gitignore

ihardz:
Resolves #225


dustinsgoodman (changes requested):
I left a comment about the upgrade and failing tests on #269 that needs to get resolved before this is merged.

dustinsgoodman:
Let's close this as a dupe of #269 then as it will be resolved there.

Issue #206, opened 2022-06-17
[cra-rxjs-sc] Show loading state in fetch example


dustinsgoodman:
# Background
There is no loading state shown on the component in case of a cold start. It would be nice to demonstrate to users how to build proper loading states. I think we should do a skeleton for the text placeholder.

# Acceptance
While the component is loading, show a loading state.

# Note
https://dev.to/jobpick/how-to-create-a-skeleton-loader-in-tailwindcss-38gh

PR #274, opened 2022-07-12
Feature/show loading state in fetch

sheerikie:
Show loading state in fetch example (Issue #206)


starters/cra-rxjs-styled-components/src/components/Loader/Loader.tsx
```diff
@@ -0,0 +1,13 @@
+import { useState } from 'react';
+
+export const Loader = () => {
+  let circleCommonClasses = 'h-2.5 w-2.5 bg-current rounded-full';
```
markshenouda, line 4: This variable isn't used

starters/cra-rxjs-styled-components/src/components/Loader/Loader.test.tsx
```diff
@@ -0,0 +1,10 @@
+import { render, screen, fireEvent } from '@testing-library/react';
```
markshenouda, line 1: fireEvent isn't used here

starters/cra-rxjs-styled-components/src/components/Loader/Loader.test.tsx
```diff
@@ -0,0 +1,10 @@
...
+import { Loader } from './Loader';
+import { MemoryRouter } from 'react-router-dom';
+
+describe('Add Loader', () => {
+  it('should appear while waiting for response', () => {
+    const { container } = render(<Loader />, { wrapper: MemoryRouter });
```
markshenouda, line 7: It's recommended to use `screen` instead in the newer versions of `@testing-library` by its author Kent C Dodds
>The benefit of using screen is you no longer need to keep the render call destructure up-to-date as you add/remove the queries you need. You only need to type screen. and let your editor's magic autocomplete take care of the rest.

> The only exception to this is if you're setting the container or baseElement which you probably should avoid doing (I honestly can't think of a legitimate use case for those options anymore and they only exist for historical reasons at this point).

starters/cra-rxjs-styled-components/src/components/Loader/Loader.test.tsx
```diff
@@ -0,0 +1,10 @@
...
+    expect(container.getElementsByClassName('w-60').length).toBe(1);
```
markshenouda, line 8: It's not recommended to select elements from the app by the Tailwind classes because they are not unique at all, you can replace it with `data-testid`

dustinsgoodman (changes requested):
Couple of things from me:

1. All of the storybook-static files probably should be added to the gitignore and not checked in.
2. CRA kit should not use tailwind.
3. The loader component is overkill for the needs of this project. It should just be a styled div or span in the FetchExample component.

PR #275, opened 2022-07-12
Override angular.json with project name vars

### LINKED ISSUE #193: [angular-apollo-tailwind] project specific naming variables need overrides

ihardz:
Resolves #193


dustinsgoodman (approved):
That seems super reasonable 😄

PR #282, opened 2022-07-21
test: test re-written

### LINKED ISSUE #280: [vue3-apollo-quasar] unit test for FetchExample component
# Background
Previous tests were written but didn't follow best practices. We want to update our tests to follow community standards and practices. Some key issues present in our tests include:

- test variables, like the component wrapper, are declared globally in a test `describe` instead of inside of `before*` or `it` scopes
- setup and teardown are done improperly
- conditionals are used in tests rather than using different `describe` scopes or unique test blocks

# Acceptance
- Move variables declarations inside of `before*` or `it` blocks; instantiation can occur inside of `describe` still
- Use setup and teardown properly to establish tests and clean up after them
- migrate conditional tests to either `it.each` blocks OR distinct `describe` blocks with their setup/teardown with the appropriate values set

# Notes 
- Read more about setup and teardown: https://jestjs.io/docs/setup-teardown
- Reducing redundant tests using `it.each`: https://dev.to/bgord/simplify-repetitive-jest-test-cases-with-test-each-310m

hdJerry:
closes #280 

Test re-written complete

### DIFF
--- a/starters/vue3-apollo-quasar/tests/unit/FetchExample.spec.ts
+++ b/starters/vue3-apollo-quasar/tests/unit/FetchExample.spec.ts
@@ -18,10 +18,14 @@ jest.mock('@vue/apollo-composable', () => {
 });
 
 describe('FetchMessage', () => {
-  const wrapper = mount(FetchExample, {
-    props: {
-      message: 'vue3-apollo-quasar starter.dev!',
-    },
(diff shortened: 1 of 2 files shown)


AllanJeremy (approved):
LGTM :shipit:

Issue #272, opened 2022-07-12
[website] update social share default copy


dustinsgoodman:
Currently, the social shares just share the link to the kit detail page. This should offer the user a default set of copy with some flavor.

# Acceptance
- Change all kit titles to "<stack> kit" (remove the word starter)
- When I click either Facebook or Twitter social share buttons, the default post should be:
> I just used the <insert kit name> by starter.dev to scaffold out my last project. It really allowed me to start building features immediately. Check out the kit: <insert link to kit>

# Notes
Example of the test for the Next kit:
> I just used the next-react-query-tailwind kit by starter.dev to scaffold out my last project. It really allowed me to start building features immediately. Check out the kit: [<insert link to kit>](https://starter.dev/kits/next-react-query-tailwind/)

kodejuice:
After researching I've discovered It is only possible with twitter, although Facebook does not allow you to change the "What's on your mind?" text box, unless of course you're developing an application for use on Facebook.

kodejuice:
We could append " Starter kit" to the kits meta description so one can tell what the link is about when shared on Facebook

![Image](https://user-images.githubusercontent.com/9017620/185386030-8efa635a-f604-46fc-86c3-29a3457ce7dd.png)

kodejuice:
@dustinsgoodman

dustinsgoodman:
We need to do a better description in general for our meta tags so that's a good shoutout. That's fine on the Facebook social bit not doing custom messages. We do need social media images and better descriptions. I'll get those into a new ticket though.

PR #301, opened 2022-08-18
feat: add default text for social share

kodejuice:
#272


kodejuice:
After researching I've discovered It is only possible with twitter, Facebook does not allow you to change the "What's on your mind?" text box, unless of course you're developing an application for use on Facebook.

We could append " Starter kit" to the kits description so one can tell what the link is about when shared on Facebook
Image

packages/website/src/components/ShareDropdown.tsx
```diff
@@ -4,14 +4,21 @@ import { ShareIcon, FacebookIcon, TwitterIcon } from '../icons';
 
 type ShareOption = 'facebook' | 'twitter';
 
-const shareUrlMap: Record<ShareOption, string> = {
-  facebook: 'https://www.facebook.com/sharer/sharer.php?u=',
-  twitter: 'https://twitter.com/intent/tweet?url=',
+const shareUrlMap: Record<ShareOption, (any) => string> = {
+  facebook: ({ u }) => `https://www.facebook.com/sharer/sharer.php?t=${u}`,
+  twitter: ({ t }) => `https://twitter.com/intent/tweet?text=${t}`,
 };
```
danecando, line 10: Can we clean these args up a little bit it just doesn't seem very clear. Maybe define a function for creating the share urls for facebook and twitter each with more clearly defined args.

I assume the changes to the twitter share no longer link to a specific URL since we aren't providing that query arg anymore. I would see if you can use both url and text for twitter and if that doesn't work just use the URL.

danecando (changes requested):

packages/website/src/components/ShareDropdown.tsx
danecando: Okay cool guess that works then

danecando (approved):
LGTM 👍

kodejuice:
resolved merge conflict

danecando (approved):

Issue #291, opened 2022-08-16
[website] Update copy


dustinsgoodman:
# Acceptance

## Nav
- [x] Update "Frameworks" in Nav to be "Kits by Tool"
  - [x] Update route to "kits-by-tool"
  - [x] Update page so routing doesn't break
- [x] "Update "GitHub Kits Library" in Nav to just be the Octocat Icon

## Homepage
- [x] Change "Tech Stacks" heading to "Available Kits"
- [x] Change "See Kit Details" on cards to "View KIt"
- [x] "Build faster. Together" -> "Build faster together"
- [x] Change copy in bottom CTA to be:
  > starter.dev is built with the community in mind. If you have ideas for improvement or new kits you'd like to see, please drop us an issue or pull request. We'd love to collaborate with you to enhance this project.
- [x] Change "Contribute" CTA to "Collaborate on starter.dev"

On "Frameworks", now "Kits by Tool"
- [x] Change Angular and Vue to be under Framework, not UI (they really are frameworks)
- [x] Change UI to "Component Library"
- [x] Change "CSS" to "Styling"

Kit page
- [x] "Download Kit" -> View Kit

PR #303, opened 2022-08-19
refactor: Update copy

kodejuice:
#291

## Nav
- [x] Update "Frameworks" in Nav to be "Kits by Tool"
  - [x] Update route to "kits-by-tool"
  - [x] Update page so routing doesn't break
- [x] "Update "GitHub Kits Library" in Nav to just be the Octocat Icon

## Homepage
- [x] Change "Tech Stacks" heading to "Available Kits"
- [x] Change "See Kit Details" on cards to "View KIt"
- [x] "Build faster. Together" -> "Build faster together"
- [x] Change copy in bottom CTA to be:
  > starter.dev is built with the community in mind. If you have ideas for improvement or new kits you'd like to see, please drop us an issue or pull request. We'd love to collaborate with you to enhance this project.
- [x] Change "Contribute" CTA to "Collaborate on starter.dev"

On "Frameworks", now "Kits by Tool"
- [x] Change Angular and Vue to be under Framework, not UI (they really are frameworks)
- [x] Change UI to "Component Library"
- [x] Change "CSS" to "Styling"

Kit page
- [x] "Download Kit" -> View Kit

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,8 +15,8 @@ Starter kits should include a set of frameworks, libraries, and essential toolin
 
 Starter kits should include some core libraries and technologies that a new project will need. These pieces should come configured according to best practices and be ready to use togther as a unit.
 
-- Framework?: (Next, Nuxt, SvelteKit, Remix, Vue, Angular, etc)
-- UI: (react)
+- Framework?: (Next, Nuxt, SvelteKit, Remix, etc)
+- Core UI: (React, Vue, Angular, etc)
 - Data Fetching and/or State Management: (redux, ngrx, rxjs, react-query, apollo, etc)
 - Styling: (tailwind, styled-components, css modules, etc)
 
--- a/README.md
+++ b/README.md
@@ -15,8 +15,8 @@ Starter kits should include a set of frameworks, libraries, and essential toolin
 
 Starter kits should include some core libraries and technologies that a new project will need. These pieces should come configured according to best practices and be ready to use togther as a unit.
 
-- Framework?: (Next, Nuxt, SvelteKit, Remix, etc)
-- UI: (react, vue, angular, etc)
+- Framework?: (Next, Nuxt, SvelteKit, Remix, Vue, Angular, etc)
+- UI: (react)
 - Data Fetching and/or State Management: (redux, ngrx, rxjs, react-query, apollo, etc)
 - Styling: (tailwind, styled-components, css modules, etc)
 
--- a/packages/website/src/config.tsx
+++ b/packages/website/src/config.tsx
@@ -76,25 +76,25 @@ export const TECHNOLOGIES = [
   {
     key: 'angular',
     name: 'Angular',
-    tags: ['Framework'],
+    tags: ['Core UI'],
     Icon: (props) => <AngularIcon {...props} />,
   },
   {
     key: 'react',
     name: 'React',
-    tags: ['Component Library'],
+    tags: ['Core UI'],
     Icon: (props) => <ReactIcon {...props} />,
   },
   {
     key: 'svelte',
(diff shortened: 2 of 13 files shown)


README.md
```diff
@@ -15,8 +15,8 @@ Starter kits should include a set of frameworks, libraries, and essential toolin
 
 Starter kits should include some core libraries and technologies that a new project will need. These pieces should come configured according to best practices and be ready to use togther as a unit.
 
-- Framework?: (Next, Nuxt, SvelteKit, Remix, etc)
-- UI: (react, vue, angular, etc)
```
danecando, line 19: You don't need to change this but I am adding my protest via comment. Maybe we need different terminology besides Framework and UI.  Maybe Meta Framework and Framework? I think vue, angular, and remix should be grouped together it's super weird that React is by itself here. cc @dustinsgoodman

kodejuice: Agree

packages/website/src/config.tsx
```diff
@@ -69,32 +69,32 @@ export const HEADER_NAV_ITEMS: NavItem[] = [
...
     key: 'svelte',
     name: 'Svelte',
-    tags: ['UI'],
+    tags: ['Component Library'],
     Icon: (props) => <SvelteIcon {...props} />,
   },
   {
     key: 'vue',
     name: 'Vue',
-    tags: ['UI'],
```
danecando, line 97: Same comment as above. Maybe Meta Framework and UI Framework?

danecando (commented):
As far as I can tell you did everything right in regards to the changes requested. I want to start a conversation about the categorization of Framework and UI though

README.md
dustinsgoodman: @danecando Yeah, I was having a lot of issues here. I think eventually we'll include Qwik, Svelte and Solid which fall in the UI Library section as they aren't fully featured frameworks. But Angular and Vue are as fully featured as Remix out of the box making this a hard debate. It's definitely a good discussion point.

danecando: I know that there is a difference using React as the example where it was always marketed as the V in MVC and just handled the UI concerns but I still think it would be easier from a user perspective to categorize the core frameworks together or even double categorize some of them. IE Angular could be part of Framework and UI Frameworks/Libraries. Or we can call them something like Meta Frameworks / Web Frameworks. Not sure what is the best solution but I think there's something that needs to be done with this

kodejuice: So whats the conclusion, should i use replace Framework with Meta Framework and UI -> Framework ? @dustinsgoodman  @danecando

dustinsgoodman: Let's go "Core UI" and "Frameworks" for now. Still debating but I think it's beyond the scope of this change to finish this debate.

danecando: cc @kodejuice

kodejuice:
@dustinsgoodman @danecando 
image

danecando (commented):
I know it's not perfect but I think people will understand it just fine. One thing I just thought of before we approve this and merge any changes - We just need to make sure that changing the name of these tags does not bring any breaking changes. IE starter kits no longer making it on the website correctly. I think they only focus on the keys but just to make sure

kodejuice:
> I know it's not perfect but I think people will understand it just fine. One thing I just thought of before we approve this and merge any changes - We just need to make sure that changing the name of these tags does not bring any breaking changes. IE starter kits no longer making it on the website correctly. I think they only focus on the keys but just to make sure

No it doesn't affect, these are just categories in this page, its not used anywhere else.

danecando (approved):

kodejuice:
fixed merge conflict

danecando (approved):

PR #318, opened 2022-10-07
Feat Nuxt 2-Pinia-Tailwind kit

jesus4497:
New kit of Nuxt 2, Pinia and Tailwind


jesus4497:
Added one extra commit adding a skeleton loader for the fetch example: 

image

Also resolved conflicts

PR #393, opened 2022-10-18
[sveltekit-pod] 1. Add linting and prettier configurations

iansamz:
Resolves: #386 

# Background

# Acceptance
- [x] Establish eslint/tslint rules and configuration
- [x] Install and configure prettier
- [x] Configure editorconfig settings
- [x] Run linting and prettier on entire project and correct any errors

# NOTE:
This task has already been completed many times on different KITs, so we should make sure to look at those configurations first and try not to reinvent the wheel :)

### DIFF
--- a/starters/svelte-kit-scss/.editorconfig
+++ b/starters/svelte-kit-scss/.editorconfig
@@ -1,0 +1,15 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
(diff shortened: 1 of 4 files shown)


asincole:
While this looks good to me, do we not need any eslint file or prettier config file in this PR? @iansamz

asincole (approved):

PR #395, opened 2022-10-18
SolidJS Linting prettier

hdJerry:
# Background

# Acceptance
- [x] Establish eslint/tslint rules and configuration
- [x] Install and configure prettier
- [x] Configure editorconfig settings
- [x] Run linting and prettier on entire project and correct any errors

# NOTE:
This task has already been completed many times on different KITs, so we should make sure to look at those configurations first and try not to reinvent the wheel :) 

Screenshot 2022-10-18 at 13 54 55
Screenshot 2022-10-18 at 13 55 26

### DIFF
--- a/starters/solidjs-tanstack-tailwind/.editorconfig
+++ b/starters/solidjs-tanstack-tailwind/.editorconfig
@@ -1,0 +1,16 @@
+# Editor configuration, see https://editorconfig.org
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+indent_size = 2
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.jsx]
+quote_type = single
+
+[*.md]
+max_line_length = off
+trim_trailing_whitespace = false
--- a/starters/solidjs-tanstack-tailwind/.editorconfig
+++ b/starters/solidjs-tanstack-tailwind/.editorconfig
@@ -9,10 +9,7 @@ insert_final_newline = true
 trim_trailing_whitespace = true
 end_of_line = lf
 
-[*.jsx]
-quote_type = single
-
-[*.js]
+[*.{js,jsx}]
 quote_type = single
 
 [*.md]
--- a/starters/solidjs-tanstack-tailwind/.editorconfig
+++ b/starters/solidjs-tanstack-tailwind/.editorconfig
@@ -7,8 +7,12 @@ indent_style = space
 indent_size = 2
 insert_final_newline = true
 trim_trailing_whitespace = true
+end_of_line = lf
 
 [*.jsx]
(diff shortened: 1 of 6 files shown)


starters/solidjs-tanstack-tailwind/.editorconfig
```diff
@@ -0,0 +1,16 @@
+# Editor configuration, see https://editorconfig.org
+root = true
+
+[*]
+charset = utf-8
```
mbicknese, line 5: editorconfig also allows us to set `end_of_line`. Shall we add that with the value of "lf"?

starters/solidjs-tanstack-tailwind/.editorconfig
```diff
@@ -0,0 +1,16 @@
...
+indent_style = space
+indent_size = 2
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.jsx]
+quote_type = single
```
mbicknese, line 12: Does this work? It doesn't seem to be in the [editorconfig spec](https://editorconfig-specification.readthedocs.io/) nor in the [Prettier API](https://prettier.io/docs/en/api.html)

starters/solidjs-tanstack-tailwind/.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
+  "trailingComma": "es5",
```
mbicknese, line 2: This is the default value and can be omitted.

mbicknese (changes requested):
Love that you added an `.editorconfig` file!

Found some small things we could touch up. If you want to keep the `quote_type` rule in there, could you also apply it to '.js' files?

starters/solidjs-tanstack-tailwind/.editorconfig
hdJerry: I think at the moment it's a proposal.
So, should we take it out?
@mbicknese 

Screenshot 2022-10-18 at 17 46 58

starters/solidjs-tanstack-tailwind/.prettierrc
hdJerry: Thanks!!

starters/solidjs-tanstack-tailwind/.editorconfig
mbicknese: That's pretty cool, let's roll with it. 👍

hdJerry: Here is the link for the screenshot.
https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties

PR #402, opened 2022-10-18
SolidJS chore: test completed

hdJerry:
# Background

Each kit should have a provided a unit testing framework fully installed and configured to be used.

# Acceptance

- [x] Add the chosen testing framework
- [x] When I run yarn test, I should see the tests pass on a simple `expect(true).toBe(true)` test.

Screenshot 2022-10-18 at 17 29 07


starters/solidjs-tanstack-tailwind/src/components/Tester/index.jsx
```diff
@@ -0,0 +1,21 @@
+import { createSignal } from 'solid-js';
+
+function Tester() {
+  const [todos, setTodos] = createSignal([]);
+
+  const toggleTodo = (value) => {
+    if (todos().includes(value)) {
+      const updateTodo = todos().filter((res) => res !== value);
+      setTodo([...updateTodo]);
+    } else if (todos().length > 0) {
```
mbicknese, line 10: Love the defensiveness, but it might not be needed here.

```javascript
let todos = [];
const value = 'newItem';
if (todos.length > 0) {
  todos = [...todos, value];
} else {
  todos = [value];
}

// todos = ['newItem']
```

has the same result as

```javascript
let todos = [];
const value = 'newItem';

todos = [...todos, value];

// todos = ['newItem']
```

starters/solidjs-tanstack-tailwind/src/components/Tester/index.jsx
```diff
@@ -0,0 +1,21 @@
+import { createSignal } from 'solid-js';
+
+function Tester() {
+  const [todos, setTodos] = createSignal([]);
+
+  const toggleTodo = (value) => {
+    if (todos().includes(value)) {
+      const updateTodo = todos().filter((res) => res !== value);
+      setTodo([...updateTodo]);
```
mbicknese, line 9:
```suggestion
      setTodos([...updateTodo]);
```

mbicknese (changes requested):
I'm not a big fan of defining new entities in `index` files. I propose we move the `Tester` component to its own `Tester.jsx` file. We can default export from there, converting to a named export in the index like so:

index.js
```javascript
export { default as Tester } from './Tester';
```

vyktoremario (commented):
This looks good. But why didn't you test the `App.jsx` file instead of creating a whole new component that will be deleted eventually.

My thought is, if a test is written for the `App.jsx`, we will just be updating the test when we update the file content.

vyktoremario:
> I did that initially but the context of the task made me just create something so that the test remains regardless of what happens to App.jsx

> When I run yarn test, I should see the tests pass on a simple expect(true).toBe(true) test. 

Based on that acceptance criteria, you just need a test file and a simple test (that is not necessarily testing the UI).

vyktoremario (approved):

mbicknese (approved):
I'm okay with having the test component for now. Seeing it's already implemented.

PR #409, opened 2022-10-19
[sveltekit-pod] 3. Create a fetch component

iansamz:
Resolves: #380 

Result: 
image

> # Background
> We want to launch each kit with an example of a component that fetches data from a data source and related tests. We have an endpoint set up as a simple hello world that accepts a greeting for example sake.
> 
> # Acceptance
> - [x] Create an API example component
> - [x] (if using REST)The component should make a GET request to `https://api.starter.dev/hello?greeting=<message>` where the message is "framework-store-style starter.dev!" 
> - [ ] (if using GraphQL) above equivalent using endpoint `https://api.starter.dev/graphql`
> - [x] The component should display the message from the API in the UI
> - [x] The component should have a storybook story
> - [x] The component should have a set of reasonable tests demonstrating how to mock APIs
> 
> # Note:
> The mock shows RxJs but that is just because it was created for a different KIT. Use the tool chosen for you current Kit
> 
> # Mock
> image


starters/svelte-kit-scss/src/routes/(examples)/api-example/+page.svelte
```diff
@@ -1,4 +1,9 @@
-<div>
-	<h1>Fetch Data From API</h1>
-	<div>Fetch component works</div>
-</div>
+<script lang="ts">
+	import Greeting from '$lib/components/Greeting/Greeting.svelte';
+
+	export let data: {
+		message: string;
+	};
```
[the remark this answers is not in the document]
iansamz: This is svelte syntax for getting data from `+page.server.ts`

starters/svelte-kit-scss/src/lib/components/Greeting/Greeting.svelte
```diff
@@ -0,0 +1,36 @@
...
+
+      h1 {
+        padding: 0.9375rem 0;
+        font-size: 2rem;
+        text-align: center;
+        border-bottom: 5px solid #1d4ed8;
+      }
+    }
+
+    div {
```
asincole, line 30: this will affect all subsequent divs, is this the intent?

starters/svelte-kit-scss/src/routes/(examples)/+layout.svelte
```diff
@@ -1,6 +1,24 @@
...
   <div>
     <a href="/">Return Home</a>
   </div>
 </div>
+
+<style lang="scss">
+  .example-container {
+    text-align: center;
+
+    div{
```
asincole, line 12: is the div in the context like the footer?
if yes, will it be better to use a class?

PR #410, opened 2022-10-19
chore: storybook added

hdJerry:
# Background
Storybook is a great way to build, validate, and test components. We want to ship a working storybook instance with each kit.

# Acceptance
- [x] Install and configure storybook
- [x] Add a simple example of storybook using the existing components

Screenshot 2022-10-19 at 17 04 04


kodejuice (approved):
LGTM

PR #419, opened 2022-10-20
Solidjs/state management

hdJerry:
# Background

Each kit should have a provided state manager setup and be configured for usage.

# Acceptance

- [x] Add the chosen state management into the Kit
- [x] make sure the store works succesfully

> We see no value in using solid-query over using SolidJS' build in createSignal, createStore and createResource primitives. We'd like to stay close to Solids native ways, and will implement the kit without tanstack (edited) 

This decision was made after a long test and assessments with the guys!!

CC: @mbicknese @vyktoremario @Zelig880 @kodejuice


mbicknese (changes requested):
It feels like you're overcomplicating what we want to deliver. If you look at the example from the Next React Tailwind starter, it is simply just the local state.
https://github.com/thisdot/starter.dev/blob/main/starters/next-react-query-tailwind/src/components/Counter/Counter.tsx

The way I see it is that other frameworks need special libraries or strategies to share state between components due to the way they're set up. SolidJS doesn't suffer from this. Signals can already safely be shared between components whilst preserving reactivity. We don't need an external store, or store-like architecture.

Additionally, the code provided does not match the screenshots in the description of this ticket. Was this deliberate?

mbicknese:
We have discovered that we have no need for this setup.

PR #420, opened 2022-10-20
[qwik-pod] feat: use graphql queries with fetch

### LINKED ISSUE #346: [qwik-pod] 1. Add store to the project
# Background

Each kit should have a provided state manager setup and be configured for usage.

# Acceptance

- [ ] Add the chosen state management into the Kit
- [ ] make sure the store works succesfully

honzikec:
Closes #346 

- Added `useQuery` that can be used to make GraphQl requests with `fetch` api.
- Refactored fetch example to use GraphQL.
- Removed some unused imports.

### DIFF
--- a/starters/qwik-graphql-tailwind/src/routes/data-fetching/data-fetching.tsx
+++ b/starters/qwik-graphql-tailwind/src/routes/data-fetching/data-fetching.tsx
@@ -1,12 +1,25 @@
 import { component$, useStore, Resource, useResource$ } from '@builder.io/qwik';
+import { useQuery } from '../../utils/useQuery';
 import * as styles from './data-fetching.classNames';
+
+export const GET_GREETING = `
+  query HelloQuery($greeting: String!) {
+    hello(greeting: $greeting)
+  }
+`;
+
+interface HelloResponse {
+  data: {
+    hello: string;
+  };
+}
 
 export const DataFetching = component$(() => {
   const store = useStore({
     greeting: '',
   });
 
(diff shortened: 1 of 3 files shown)


starters/qwik-graphql-tailwind/src/routes/data-fetching/data-fetching.tsx
```diff
@@ -34,16 +47,27 @@ export const DataFetching = component$(() => {
...
 });
 
-export async function fetchGreeting(greeting: string, abortController?: AbortController): Promise<string> {
-  const resp = await fetch(`https://api.starter.dev/hello?greeting=${greeting}`, {
+export async function fetchGreeting(greeting: string, abortController?: AbortController): Promise<HelloResponse> {
+  if (!greeting) {
+    greeting = 'there';
+  }
+
+  const { executeQuery$ } = useQuery(GET_GREETING);
```
JiaLiPassion, line 62: Just a question here, the `useQuery` looks great, and I am just curious, since the `fn` passed to `useResource$` will be transformed via `$` and become lazy loaded trunk, in this case `fetchGreeting` will become a `$` function, so do we need to optimize the `executeQuery$` too?

honzikec: Good question! I would say indeed in this case we don't really need that optimization, but it doesn't hurt to keep it generic in case we'd want to use the `useQuery` from a different context....

SDaian (approved):
LGTM

PR #449, opened 2022-10-24
Solidjs/add kit cli

hdJerry:
- [x] Add the kit to the CLI as a setup option


mbicknese (changes requested):
Can you remove the two commits that do not belong here? It's hard to review the changes when non-relevant code is mixed in.

PR #465, opened 2022-10-26
fix(solidjs): replace placeholder <kit> tag

mbicknese:
the `<kit name>` broke the markdown parser 🤦


vyktoremario (approved):
Tried reading it as an HTML tag.👍

hdJerry (approved):
fair enough. LOL

PR #472, opened 2022-10-28
fix(#464): hide showcase links for kits with no showcase

### LINKED ISSUE #464: Showcase should be hidden if there isn't a showcase
Image

jnalei:
This PR closes #464 

"View Showcase" links were previously visible on kits that did not have a showcases. 
Links are now properly rendered based on `hasShowcase` value set in each kit's `package.json` file.

In the example picture below, `qwik-tailwind-graphql` kit does not have a showcase so the link does not display.
Screen Shot 2022-10-27 at 10 05 02 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ For proper integration with the starter-dev CLI and website there are also some 
 
 - `description` - This field is used as a short description highlighting the technologies of each particular starter kit. It is shown as a selection in the starter-dev CLI. Example: `NextJS, React Query, and TailwindCSS`
 - `keywords` - This field is an array of strings used to map to the technology list on the starter.dev website. Example: `["next", "react-query", "tailwind"]`
+- `hasShowcase` - This field is a boolean and is required when displaying the "View Showcase" link for kits that have showcases. If your kit has a showcase, set the value to `true`. If your kit does not have a showcase, set the value to `false`.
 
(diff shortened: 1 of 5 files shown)


jdwilkin4 (dismissed):
I tested this out locally and everything is working on my end. 
The only thing I think is missing is letting future contributors know that they should add the `hasShowcase` key in the `package.json`. I can see developers forgetting to add that and for ones that have a showcase it won't show without that key set to true. 

I don't know if we want to add something in the readme or a checklist item in a PR template(which we should probably have anyway) but that is something for us to think about in the future. 👍

jnalei:
> I tested this out locally and everything is working on my end. The only thing I think is missing is letting future contributors know that they should add the `hasShowcase` key in the `package.json`. I can see developers forgetting to add that and for ones that have a showcase it won't show without that key set to true.
> 
> I don't know if we want to add something in the readme or a checklist item in a PR template(which we should probably have anyway) but that is something for us to think about in the future. 👍

@jdwilkin4 Yes! Good point and thank you for reminding me! I meant to add a note in the readme then completely forgot about it.. will update in another commit 👍

jdwilkin4 (approved):
LGTM 👍

morgnism (approved):
LGTM

PR #486, opened 2022-11-04
chore(next12-react-query-tailwind): Rename kit, upgrade dependencies, improvements

danecando:
This PR gets the `next-react-query-tailwind` up to date with its dependencies, configurations, and conventions. It also renames the kit to `next12-react-query-tailwind` to differentiate between the potential next 12 and 13 starter kits since they are vastly different frameworks. It also includes some random fixes and improvements.

### Changes
- renames `next-react-query-tailwind` starter kit to `next12-react-query-tailwind`
- upgrades all of the starter kit dependencies
- updates to use new next testing setup utilities and guidelines
- adds vscode config `eslint.workingDirectories` so that vscode will register each project in the monorepo as it's own project to avoid conflicts and issues with eslint and other tools
- other random small fixes and improvements


starters/next12-react-query-tailwind/package.json
```diff
@@ -0,0 +1,63 @@
+{
+  "name": "next12-react-query-tailwind",
+  "description": "NextJS, React Query, and TailwindCSS",
```
jdwilkin4, line 3: it looks like this one needs to be renamed too

starters/next12-react-query-tailwind/package.json
```diff
@@ -0,0 +1,63 @@
...
+  "keywords": [
+    "next",
```
jdwilkin4, line 5: Do we need to create two keywords? 
one for next12 and another for next13?

dustinsgoodman: Should be nextjs. package.json keywords are for npm search and the versioning doesn't quite matter there.

danecando: It also maps to technologies on the website so it has to match up with that and we don't want different next js coming up

starters/next12-react-query-tailwind/README.md
```diff
@@ -1,4 +1,4 @@
-# next-react-query-tailwind starter kit
+# next12-react-query-tailwind starter kit
 
 This starter kit features **Next.js**, **React Query**, and **Tailwind CSS**.
```
jdwilkin4, line 3: I think this line should be change too so it says Next.js v.12 or something like that.

jdwilkin4 (changes requested):
I think it looks good. I just added one place where I think the name should be updated to Next 12 👍

PR #524, opened 2022-11-14
[GraphQL-Serverless-Contentful] Init the starter

markshenouda:
# What in this PR?
- A minimal example to start serverless function offline
- Installation of static testing analysis tools
   - TypeScript
   - Prettier
   - ESlint
  
First serverless function


starters/graphql-serverless-contentful/.build/package.json
```diff
@@ -0,0 +1 @@
+/Users/markshenouda/Projects/first-serverless/package.json
```
jdwilkin4, line 1: I feel like we should remove this line because it mentions your personal user account

jdwilkin4 (changes requested):
For the `.prettierrc` file, we should add some basic rules there unless it will be addressed in other PR

starters/graphql-serverless-contentful/.build/package.json
markshenouda: Oh this `.build` folder shouldn't be here

jdwilkin4 (approved):

PR #528, opened 2022-11-17
fix(#476): add support for removing unwanted fields in package.json for generated kits

### LINKED ISSUE #476: Add support for removing package.json fields from generated starter kits
We added a new field 'hasShowcase' and I realized that this will be an odd random field that ends up in starter kit users package.json files. Would be good to automatically remove any fields like this from the generated kits

jnalei:
## Summary
When kits are generated by the user, the `hasShowcase` field also shows up in the `package.json` file. This field is needed in the starter kits but unnecessary for kits generated by the user. This PR provides a solution to exclude the unwanted field. 

## Type of change

- [ ] Documentation change
- [x] Bug fix

## Checklist

- [x] This fix resolves #476 

Closes #476


TapaiBalazs (approved):
LGTM

PR #573, opened 2022-12-01
[NODE] feat: added customised logger to the application

### LINKED ISSUE #550: [node-pod] Set Up A Generic Logger
Make a logger class, that puts the proper prefixes to error/warn/info logs

e.g.: `[ERROR] this is an error log and the [ERROR] is attached it when it gets called`

`Logger.log() Logger.error()`

# Notes
- Needs to follow best practices in making log entries easy to query in log helper tools (cloudwatch, splunkr, etc)

TapaiBalazs:
## Type of change

- [ ] Documentation change
- [ ] Bug fix
- [X] Feature

## Summary of change

- Created a customised logger that is configurable by log levels
- Let's have a discussion about adding morgan/pinojs to the starter

## Checklist

- [ ] This fix resolves #550


starters/express-typescript-typeorm-postgres/src/utils/log-helper.ts
```diff
@@ -0,0 +1,59 @@
...
+    LOG_LEVELS = parsedLogLevels;
+  }
+} catch (e) {
+  console.warn(
+    `[WARN] Could not parse ALLOWED_LOG_LEVELS env variable, logger defaults to all levels`,
+    e
+  );
+}
+
+class Logger {
```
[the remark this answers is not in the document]
TapaiBalazs: No, the class is instantiated only once and that one instance gets exported as a singleton.

starters/express-typescript-typeorm-postgres/src/utils/log-helper.ts
```diff
@@ -0,0 +1,59 @@
...
+  private enabledLevels: Set<LogLevel>;
+
+  constructor(allowedLogLevels: LogLevel[]) {
+    this.enabledLevels = new Set(allowedLogLevels);
+    this.log(`Initialising logger with allowed levels: ${allowedLogLevels.join(', ')}`);
+  }
+
+  log(...args) {
```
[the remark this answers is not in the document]
TapaiBalazs: I updated these to mimic the console method types.

starters/express-typescript-typeorm-postgres/src/utils/log-helper.ts
```diff
@@ -0,0 +1,72 @@
...
+    if (this.enabledLevels.has('DEBUG')) {
+      if (typeof message === 'string') {
+        console.debug(`[DEBUG] ${message}`, ...optionalParams);
+      } else {
+        console.debug(`[DEBUG]`, message, ...optionalParams);
+      }
+    }
+  }
+
+  warn(message?: any, ...optionalParams: any[]): void {
```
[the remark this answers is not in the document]
TapaiBalazs: These are the types of the console methods, we only have a wrapper around the console, therefore, these methods should have the same types as the console methods.

starters/express-typescript-typeorm-postgres/src/utils/log-helper.ts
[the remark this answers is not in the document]
TapaiBalazs: No it isn't.

starters/express-typescript-typeorm-postgres/src/utils/log-helper.ts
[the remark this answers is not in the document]
TapaiBalazs: Because it is a log helper which allows setting the log levels and uses console under the hood. This way we import the helper which has methods that developers are used to, and it adds extra value when looking for log entries in a tool like cloudWatch

starters/express-typescript-typeorm-postgres/.env.example
```diff
@@ -2,11 +2,14 @@
 
 NODE_ENV=development
 
+ALLOWED_LOG_LEVELS=["DEBUG","INFO","WARN","ERROR"]
```
BigAB, line 5: JSON in env files, that's unusual.

starters/express-typescript-typeorm-postgres/src/utils/log-helper.ts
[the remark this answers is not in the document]
BigAB: I don't see the point of creating a class here.

You could just do everything in the module and export `log`, `info`, `warn`, ..etc directly from the module

starters/express-typescript-typeorm-postgres/src/utils/log-helper.ts
```diff
@@ -0,0 +1,72 @@
...
+      if (typeof message === 'string') {
+        console.error(`[ERROR] ${message}`, ...optionalParams);
+      } else {
+        console.error(`[ERROR]`, message, ...optionalParams);
+      }
+    }
+  }
+}
+
+export const LogHelper = new Logger(LOG_LEVELS);
```
BigAB, line 72: I don't think we should export this as `LogHelper`, just `Logger` or better yet just export the individual log methods.

starters/express-typescript-typeorm-postgres/src/utils/log-helper.ts
```diff
@@ -0,0 +1,72 @@
...
+    `[WARN] Could not parse ALLOWED_LOG_LEVELS env variable, logger defaults to all levels`,
+    e
+  );
+}
+
+class Logger {
+  private enabledLevels: Set<LogLevel>;
+
+  constructor(allowedLogLevels: LogLevel[]) {
+    this.enabledLevels = new Set(allowedLogLevels);
```
BigAB, line 23: One concern to note, which decided how succinct the code can be, is whether or not we want to support the case where `process.env.ALLOWED_LOG_LEVELS` changing after startup. Assuming we do we, we should probably check the `process.env.ALLOWED_LOG_LEVELS` every time.

starters/express-typescript-typeorm-postgres/src/utils/log-helper.ts
```diff
@@ -0,0 +1,72 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+const allowedLogLevels = process.env.ALLOWED_LOG_LEVELS || '[]';
+
+type LogLevel = 'INFO' | 'DEBUG' | 'WARN' | 'ERROR';
+
+let LOG_LEVELS: LogLevel[] = ['DEBUG', 'INFO', 'WARN', 'ERROR'];
+try {
+  const parsedLogLevels = JSON.parse(allowedLogLevels);
```
BigAB, line 8: We probably don't need to parse this into an object, we could probably just check `process.env.ALLOWED_LOG_LEVELS` for the strings, then people could do comma separated or whatever
```ts
new RegExp(LOG_LEVEL).test(process.env.ALLOWED_LOG_LEVELS)
```

starters/express-typescript-typeorm-postgres/src/utils/log-helper.ts
```diff
@@ -0,0 +1,72 @@
...
+    if (this.enabledLevels.has('ERROR')) {
+      if (typeof message === 'string') {
+        console.error(`[ERROR] ${message}`, ...optionalParams);
+      } else {
+        console.error(`[ERROR]`, message, ...optionalParams);
+      }
+    }
+  }
+}
+
```
BigAB, line 71: We probably also need to address the need to replace the default logger with a custom logger of some sort.

Maybe an interface like this:
```ts
interface Logger {
  log: (...args: unknown[]) => void;
  info: (...args: unknown[]) => void;
  debug: (...args: unknown[]) => void;
  warn: (...args: unknown[]) => void;
  error: (...args: unknown[]) => void;
}

export const setCustomLogger = (customLogger: Logger) => {/*...*/}
```

starters/express-typescript-typeorm-postgres/src/utils/log-helper.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+  let consoleSpy;
+  beforeEach(() => {
+    jest.resetModules();
+  });
+
+  afterEach(() => {
+    jest.resetAllMocks();
+  });
+
+  describe(`log method`, () => {
```
BigAB, line 11: The tests are very repetitive we could simplify it using the `describe.each()` feature
```ts
describe.each([
      ['log', 'INFO'],
      ['info', 'INFO'],
      ['warn', 'WARN'],
      ['debug', 'DEBUG'],
      ['error', 'ERROR'],
    ])('%s function', (method, level) => {
      test(`calls logger (console by default) if ${level} is set in process.env.ALLOWED_LOG_LEVELS`, () => {
        process.env.ALLOWED_LOG_LEVELS = 'INFO,WARN,DEBUG,ERROR';
        logger[method]('Test');
        expect(console[method]).toHaveBeenCalledWith(`[${level}]`, 'Test');
      });

      test(`does NOT call logger (console by default) if level is NOT in process.env.ALLOWED_LOG_LEVELS`, () => {
        process.env.ALLOWED_LOG_LEVELS = '';
        logger[method]('Test');
        expect(console[method]).not.toHaveBeenCalled();
      });
    });
```

starters/express-typescript-typeorm-postgres/src/utils/log-helper.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+    beforeEach(() => {
+      consoleSpy = jest.spyOn(console, 'log');
+    });
+
+    it(`does not call console.log`, () => {
+      process.env.ALLOWED_LOG_LEVELS = `["ERROR"]`;
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      const { LogHelper } = require('./log-helper');
```
BigAB, line 19: We should probably use `await import('./log-helper')` instead of `require()` for consistency and a modern approach

BigAB (commented):
I don't think we should call it a `helper` or put it in `util`.
I think we should just put it in a directory called `logger` or something.

ALSO: This could be a lot more succinct.

I'm thinking something like 
```ts
type LogLevel = 'DEBUG' | 'INFO' | 'WARN' | 'ERROR';

const logger = (method, level: LogLevel = 'INFO') => {
  const levelRegExp = new RegExp(level);
  return (message?: unknown, ...optionalParams: unknown[]) => {
    if (levelRegExp.test(process.env.ALLOWED_LOG_LEVELS)) {
      console[method](`[${level}]`, message, ...optionalParams);
    }
  };
};

export const log = logger('log');
export const info = logger('info');
export const debug = logger('debug', 'DEBUG');
export const warn = logger('warn', 'WARN');
export const error = logger('error', 'ERROR');
```

Sorry about a lot of comments here, but I do think there can be improvements

starters/express-typescript-typeorm-postgres/src/utils/log-helper.ts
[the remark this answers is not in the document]
TapaiBalazs: I really want these methods to have the exact same type as the console methods, and I linked before the official types of the console methods.

starters/express-typescript-typeorm-postgres/src/utils/log-helper.ts
TapaiBalazs: I initially went with `Logger` but one of our devDepencencies (concurrenly) has a Logger exported and I don't want people to accidentally import that instead of our customised one. I would like to keep the `LogSomething.log()` syntax to indicate that it is a wrapper around the console. I'm open for naming suggestions, or do you think it would worth the risk of naming it Logger and maybe cause users to accidentally import (or auto-import) the wrong one? 🤔

starters/express-typescript-typeorm-postgres/src/utils/log-helper.ts
TapaiBalazs: I thought about it, and my recent experience with logging and bringing down a whole production AWS infrastructure by adding more and more logic to logging says that we should not allow that, but if somebody needs that functionality in their app, they can certainly implement it building on top of our code.

**My reasoning:**
Even if we simplify the environment variables to be a comma separated list (I'll get to it soon), it is two or more operations:
1. Read the env variable
2. Run `string.prototype.split()` on it
3. Iterate over the array to find if the wanted log-level is in there

Doing these 3 steps on every log has a toll on performance, even if only "Error" is allowed, because the above 3 steps will run on every log call no matter what. So a production code of:

```
function somethingHandler(req, res, next) {
  Logger.debug(req.url)
  let something = req.params.something
  if (!something) {
    Logger.warn(`${req.params.something} is undefined, method was called from an app below version 3.2. Defaulting to "whatever" value`
    something = "whatever
  }
  // ...
  
}
```

Would do those operations even if we don't log anything because of the log levels. That is why I'd avoid putting too much logic into each logger call.

starters/express-typescript-typeorm-postgres/src/utils/log-helper.ts
TapaiBalazs: I recently brought down a production environment with using regexes in logging. I'd feel safer with going `.split(',')` and putting that into a Set.

starters/express-typescript-typeorm-postgres/src/utils/log-helper.ts
TapaiBalazs: Do you want to monkey-patch the console object in this method? 🤔 

```
const oldLogMethod = console.log;
console.log = Logger.log
```

starters/express-typescript-typeorm-postgres/src/utils/log-helper.spec.ts
TapaiBalazs: I am generally not fond of these types of test abstractions. I know we could do that, but whenever something later gets customised it is a pain to fix these abstractions.

That said, I'll modify the tests as per your suggestions, because it is not likely these are going to be broken later.

starters/express-typescript-typeorm-postgres/src/utils/log-helper.ts
```diff
@@ -0,0 +1,72 @@
...
+    this.enabledLevels = new Set(allowedLogLevels);
+    this.log(`Initialising logger with allowed levels: ${allowedLogLevels.join(', ')}`);
+  }
+
+  log(message?: any, ...optionalParams: any[]): void {
+    if (this.enabledLevels.has('INFO')) {
+      if (typeof message === 'string') {
+        console.log(`[INFO] ${message}`, ...optionalParams);
+      } else {
+        console.log(`[INFO]`, message, ...optionalParams);
```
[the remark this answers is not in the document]
TapaiBalazs: ok

starters/express-typescript-typeorm-postgres/src/utils/log-helper.spec.ts
TapaiBalazs: ok

starters/express-typescript-typeorm-postgres/src/utils/log-helper.ts
TapaiBalazs: We use concurrently as a devDependency, but since it is in the node_modules folder, almost all import/autoImport tools will default to concurrently because of it's alphabetical precedence.See screenshot

Screenshot 2022-12-05 at 10 27 09

starters/express-typescript-typeorm-postgres/src/utils/log-helper.ts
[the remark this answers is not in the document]
TapaiBalazs: @BigAB my problem with that approach is similar as to naming the thing as `Logger`, namely some of the functions would have multiple imports, mostly from dev dependencies:

Screenshot 2022-12-05 at 10 32 15

starters/express-typescript-typeorm-postgres/src/utils/log-helper.ts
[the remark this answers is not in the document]
BigAB: `any` turns off  type checking, but `unknown` does not.

I am thinking hard, but I can't think of an instance where having type `unknown` would behave any differently that having `any` except it would allow us to stay in `strict` mode for TS.

Can you think of anything that may possibly be different or make something weird if we used `unknown`?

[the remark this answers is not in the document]
TapaiBalazs: I implemented your unknown approach, thanks for the suggestion. :)

starters/express-typescript-typeorm-postgres/src/utils/log-helper.ts
TapaiBalazs: We leave it as LogHelper for now, and we'll rename it if we can find a better name that won't try to import from dev dependencies.

starters/express-typescript-typeorm-postgres/src/utils/log-helper.ts
TapaiBalazs: We discussed this, we just want to provide a basic wrapper, but we don't want to replace the logging implementation from console. If users want their own logger, they can use any third-party logger implementation that supports that. (e.g.: winston, etc)

BigAB (approved):

PR #592, opened 2022-12-05
[Deno] Generate Project

vyktoremario:
## Type of change

- [ ] Documentation change
- [ ] Bug fix
- [x] New Kit setup

## Checklist

- [ ] This fix resolves #


dariodjuric (changes requested):
I would also suggest to add a lock file that locks the Oak version.

https://deno.land/manual@v1.28.2/basics/modules/integrity_checking

starters/deno-oak-denodb/src/main_test.ts
```diff
@@ -0,0 +1,5 @@
+import { assertEquals, testing } from '../dev_deps.ts';
+
+Deno.test(function addTest() {
```
dariodjuric, line 3: Do we need this test or can we hold off with dummy tests until we have something real to test? 🤔

dariodjuric: I was just suggesting to remove it as it doesn't provide any value, but if you think we should keep it, let's keep it.

vyktoremario: I can remove it then.

vyktoremario: Done

dariodjuric (approved):

Issue #605, opened 2022-12-06
[deno] Add database migration


dariodjuric:
Given [this database schema](https://dbdiagram.io/d/638f24e3bae3ed7c45450a5f), we need to perform schema migration as well as initial seed of the database.

### Acceptance Criteria

- [x] Create a database schema for https://dbdiagram.io/d/638f24e3bae3ed7c45450a5f
- [ ] Document the command required to execute the database schema
- [x] Create some initial seed data (e.g. 3 users)
- [ ] Document the command required to seed the database
- [x] Create a GraphQL API
- [x] The API should have CRUD operations for users. See the database model: https://dbdiagram.io/d/638f24e3bae3ed7c45450a5f
- [ ] Nice to have: one command to run both the database container and the web server

PR #610, opened 2022-12-06
[Deno] Database migration

vyktoremario:
## Type of change

- [ ] Documentation change
- [ ] Bug fix
- [x]  Feature

## Summary of change

Closes: #605 

## Checklist

- [ ] This fix resolves #


starters/deno-oak-denodb/src/db/resolvers/technology.ts
```diff
@@ -0,0 +1,51 @@
+import { Technologies } from '../model/technology.ts';
+
+export interface ITechnology {
+	displayName: string;
+	description: string;
+	url: string;
+}
+
+export const technologyResolvers = {
+	Query: {
```
dariodjuric, line 10: Do you think we could put query handlers to a separate module, as well as mutations to its own file?

starters/deno-oak-denodb/src/main.ts
```diff
@@ -1,24 +1,38 @@
...
+	}),
 );
+
+app.use(GraphQLService.routes(), GraphQLService.allowedMethods());
 app.use(router.routes());
 app.use(router.allowedMethods());
 
+await db.sync({ drop: true });
+
+console.log(`Database connected!`);
```
dariodjuric, line 35: What do you think about logging the hostname of the database as well?

starters/deno-oak-denodb/src/main.ts
```diff
@@ -1,24 +1,38 @@
...
+		origin: CORS_ALLOWED_ORIGINS ? CORS_ALLOWED_ORIGINS.split(',') : '',
+		optionsSuccessStatus: 200, // Some legacy browsers (IE11, various SmartTVs) choke on 204
+	}),
 );
+
+app.use(GraphQLService.routes(), GraphQLService.allowedMethods());
 app.use(router.routes());
 app.use(router.allowedMethods());
 
+await db.sync({ drop: true });
```
dariodjuric, line 33: This is a dangerous operation to be in the app startup module, especially with `drop: true`. I would suggest to place it in its own module.

starters/deno-oak-denodb/src/db/resolvers/technology.ts
vyktoremario: I have moved it, but I am not sure if that's what you mean

starters/deno-oak-denodb/src/main.ts
dariodjuric: For example, you can place it in `seed.ts` and then the seed command would be `deno run seed.ts ...`. It's a manual command that needs to be executed during project setup.

starters/deno-oak-denodb/src/db/model/technology.ts
```diff
@@ -0,0 +1,16 @@
+import { DataTypes, Model } from '../../../deps.ts';
+
+export class Technologies extends Model {
```
dariodjuric, line 3: Shouldn't we be using a `User` model here?

starters/deno-oak-denodb/src/main.ts
dariodjuric: Oh, I see you already have `run_seeder.ts`. Yeah, that's good.

So you can remove this line then.

starters/deno-oak-denodb/src/db/model/technology.ts
dariodjuric: Or if we have Technologies in other starter kit, we can use that, sure.

starters/deno-oak-denodb/src/db/run_seeder.ts
```diff
@@ -0,0 +1,30 @@
...
+import { technologySeedData } from './seeding/technology-seeder.ts';
+
+import {
+	DATABASE_HOST,
+	DATABASE_NAME,
+	DATABASE_PASSWORD,
+	DATABASE_USERNAME,
+} from '../config/environment.ts';
+
+async function runSeeder() {
```
dariodjuric, line 12: Could we add an instruction in the readme on how to run the seeder?

starters/deno-oak-denodb/src/db/resolvers/mutationHandler.ts
```diff
@@ -0,0 +1,38 @@
+import { Technologies } from '../model/technology.ts';
+export interface ITechnology {
```
dariodjuric, line 2: I'm not clear on why this interface is needed. Should it correspond to the GraphQL type `Technology`?

dariodjuric (changes requested):
I left some questions/comments.

starters/deno-oak-denodb/src/db/model/technology.ts
vyktoremario: Based on the Document Dustin shared, this was the model he said the backend kits should have

starters/deno-oak-denodb/src/db/resolvers/mutationHandler.ts
vyktoremario: This corresponds to the `TechnologyInput` in the type definitions.  
I can use `any` for now

starters/deno-oak-denodb/src/db/model/technology.ts
dariodjuric: Ah, okay. I missed that. 🙊

starters/deno-oak-denodb/src/db/resolvers/mutationHandler.ts
dariodjuric: I see. In that case, I would suggest to name it that way: `TechnologyInput`. You could also place it in its own folder under the GraphQL API folder, e.g. `types`.

Which reminds me, ne other thing that we might consider is splitting the GraphQL-related code into its own folder. Right now it's located under the `db` folder, but the GraphQL code is really database-independent. What do you think about having a `graphql` folder, which would host API types like `TechnologyInput`, the schema, resolvers, etc.

dariodjuric: Can we generate those somehow?

If so, we should have this documented in the readme on how to do it. The initial types should be generated by us.

vyktoremario: For now, let's leave things as they are. 
I will be trying that in a different PR

dariodjuric (approved):

PR #613, opened 2022-12-07
[NODE] Feat/595 add cache to starter kit

### LINKED ISSUE #595: [node-pod] Adding Cache To The Node Pod

TapaiBalazs:
## Type of change

- [ ] Documentation change
- [ ] Bug fix
- [x] Feature

## Summary of change

- Moved docker setup to docker-compose setup
- Added database connect retry for startup to mitigate slow database spin-up
- Added a redis instance to our infrastructure
- Added `cachified` package
- Updated Example handlers to use caching
- Updated unit tests
- Added redis to health-check

## Checklist

- [x] This fix resolves #595 


starters/express-typescript-typeorm-postgres/src/db/datasource.ts
```diff
@@ -12,6 +11,33 @@ export const dataSource = new DataSource({
...
   entities: ['**/*.entity.js'],
-  logging: IS_DATABASE_LOGGING_ENABLED,
-  synchronize: true,
+  logging: DATABASE_ENABLE_LOGGING,
+  synchronize: DATABASE_ENABLE_SYNC,
 });
+
+const DATABASE_RETRY_COUNT = process.env.DATABASE_RETRY_COUNT
+  ? parseInt(process.env.DATABASE_RETRY_COUNT)
+  : 25;
```
BigAB, line 20: `25` retries by default?  Isn't that a lot?

starters/express-typescript-typeorm-postgres/src/db/datasource.ts
```diff
@@ -12,6 +11,33 @@ export const dataSource = new DataSource({
...
+    .initialize()
+    .then(() => true)
+    .catch((err) => {
+      const remainingRetries = retries - 1;
+      console.warn(`Could not connect to the database, retrying ${remainingRetries} more time(s)`);
+      if (remainingRetries === 0) {
+        console.error(`Error during Data Source initialisation:`, err);
+        return false;
+      }
+      return new Promise((resolve) => {
```
BigAB, line 37: I think you will need to handle the reject here as well, no?
```ts
return new Promise((resolve, reject) => {
  setTimeout(() => {
    initialiseDataSource(remainingRetries).then(resolve, reject)
  }, DATABASE_RETRY_INTERVAL_MS);
});
```

starters/express-typescript-typeorm-postgres/src/controllers/health/handlers/get-health.ts
```diff
@@ -1,12 +1,16 @@
 import { Request, Response } from 'express';
 import { getReasonPhrase, StatusCodes } from 'http-status-codes';
-import { dataSource } from '../../../db/datasource';
+import { cacheRedisClient, redisHealthCheck } from '../../../cache/cache';
```
BigAB, line 3: > error TS2459: Module `"../../../cache/cache"` declares `cacheRedisClient` locally, but it is not exported.

> error TS2305: Module `"../../../cache/cache"` has no exported member `redisHealthCheck`

starters/express-typescript-typeorm-postgres/docker-compose.yml
```diff
@@ -0,0 +1,23 @@
...
+
+  redis_cache:
+    image: 'redis:alpine'
+    container_name: 'cache'
+    command: redis-server
+    ports:
+      - '6379:6379'
+    volumes:
+      - ./cache_data:/var/lib/redis
+      - ./cache_conf:/usr/local/etc/redis/redis.conf
```
BigAB, line 21: I had a lot of trouble with this.

When I ran `npm run dev` the first time, it seemed to work. 👍 

The seconds time, I got a bunch of errors:
```sh
[INFO] Example app listening on port 3333
[ERROR] [CACHE] An error occurred while connecting to Redis Error: connect ECONNREFUSED 127.0.0.1:6379
    at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1157:16) {
  errno: -61,
  code: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 6379
}
Could not connect to the database, retrying 24 more time(s)
[INFO] [CACHE] Reconnecting to Redis instance
[ERROR] [CACHE] An error occurred while connecting to Redis Error: connect ECONNREFUSED 127.0.0.1:6379
    at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1157:16) {
  errno: -61,
  code: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 6379
}
[INFO] [CACHE] Reconnecting to Redis instance
[ERROR] [CACHE] An error occurred while connecting to Redis Error: connect ECONNREFUSED 127.0.0.1:6379
    at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1157:16) {
  errno: -61,
  code: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 6379
}
```

Then the next time, after looking in the docker dashboard, I got this weirdness:
```sh
[INFO] Example app listening on port 3333
[INFO] [CACHE] Successfully connected to Redis instance
[DEBUG] Connected to redis
Could not connect to the database, retrying 24 more time(s)
Could not connect to the database, retrying 23 more time(s)
Could not connect to the database, retrying 22 more time(s)
Could not connect to the database, retrying 21 more time(s)
Could not connect to the database, retrying 20 more time(s)
```

So, something is not right here. Is there maybe a race condition or something, the docker dashboard log said it couldn't find a config.

BigAB (commented):
This all looks pretty good, I've left some comments and questions.

Here's one just to think about, not necessarily implement. What would this caching module look like if it was implemented as an express middleware?

I wonder if we could check the cache based on the request and return the result, skipping the rest of the middlewares/routers. And conversely looking at the response after the router handlers have run an caching the result.

I wonder if that would be a good idea, then the individual controllers do not need to know anything about the cache (or maybe they would still need to know which bits to invalidate on update if we can't generalize that behaviour).

Just a thought

starters/express-typescript-typeorm-postgres/README.md
```diff
@@ -56,7 +57,9 @@ yarn create @this-dot/starter --kit express-typescript-typeorm-postgres
 
 - Follow the prompts to select the `express-typescript-typeorm-postgres` starter kit and name your new project.
 - `cd` into your project directory and run `npm install`.
-- Run `npm run dev` to start the development server.
+- Create a `.env` file and copy the contents of `.env.example` into it.
+- Run `npm run dev` to start the development server and infrastructure.
+- Make sure you have docker & docker-compose installed on your machine
```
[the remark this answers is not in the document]
TapaiBalazs: Right, I updated the readme accordingly

starters/express-typescript-typeorm-postgres/src/controllers/health/handlers/get-health.ts
```diff
@@ -16,26 +20,54 @@ export async function getHealth(req: Request, res: Response): Promise<void> {
     return;
   }
 
+  if (redisPingResult.type === Result.ERROR) {
+    /**
+     *  We return 206 here, because if the cache server is down the application is still working,
```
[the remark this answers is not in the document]
TapaiBalazs: That is an excellent question and I did not find a good answer with Simone. Health checks are usually used for spinning up or down instances. This scenario is when the API can serve the clients, but parts of it are not working, that is why I thought 206 - Partial would be good here, but I'm not even convinced.

My reasoning behind a different status code is that as far as I know we can set up alerts some monitoring tools for certain status codes that calls the health endpoint regularly and can send out alerts to on-call personnel, etc.

I am open to suggestion though. According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#successful_responses), 207 is Multi-Status

> 207 Multi-Status
> Conveys information about multiple resources, for situations where multiple status codes might be appropriate.

Do you think that's more appropriate? 🤔

starters/express-typescript-typeorm-postgres/src/controllers/technology/handlers/delete-technology.ts
```diff
@@ -1,22 +1,49 @@
...
+
+  clearCacheEntry(req.baseUrl);
+  clearCacheEntry(req.originalUrl);
+
+  res.status(StatusCodes.OK).json({
+    id: technologyId,
+  });
+}
+
+function deleteTechnologyEntry(technologyId: number): Promise<DeleteTechnologyResult> {
```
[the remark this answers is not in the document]
TapaiBalazs: I agree on extracting the methods, I'm not sure about abstracting it into a standard interface. Do we want to be that opinionated? Let's get back to this when you are online.

starters/express-typescript-typeorm-postgres/src/controllers/technology/handlers/get-technology.ts
```diff
@@ -1,50 +1,63 @@
...
-        error: getReasonPhrase(StatusCodes.NOT_FOUND),
-        details: technologyResult.message,
-      });
-      return;
-    }
-    res.json(technologyResult.data);
-  } catch (e) {
-    next(e);
+  if (technologyResult.type === Result.NOT_FOUND) {
+    res.status(StatusCodes.NOT_FOUND).json({
```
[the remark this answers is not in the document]
TapaiBalazs: My reasoning behind creating a new Result for `NOT_FOUND` is that in this scenario we have the second if under the first one, and I think that is more readable.

My problem of having different error types is that it would then need if cases inside an if case, which increases cognitive complexity

Who knows maybe in the future there are going to be more result types, like BAD_REQUEST, or FORBIDDEN. You know I'm not sure either anymore. 🤔

starters/express-typescript-typeorm-postgres/src/db/datasource.ts
TapaiBalazs: I was testing something and forgot it there, I updated it to be .env based and updated the .env.example file accordingly.

starters/express-typescript-typeorm-postgres/src/db/datasource.ts
TapaiBalazs: I don't think so, because it is recursion, and the `initialiseDataSource` method calls the `initialiseDataSource` method which always resolves.
In the `.catch()` callback it returns early, when the `remainingRetries === 0` condition is true and AFAIK returning and not throwing in a catch resolves as a promise.

starters/express-typescript-typeorm-postgres/src/controllers/health/handlers/get-health.ts
TapaiBalazs: Sorry about that, my IDE stopped renaming imports when I move files around. I'll investigate what happened with it, it was a feature I heavily relied upon and now it doesn't work :(

starters/express-typescript-typeorm-postgres/docker-compose.yml
TapaiBalazs: That's strange. 🤔 I'm afraid we will have to look into it together, because when I do an `npm run infrastructure:clear` and then start it up again it works for me. The database does need polling for the first init, because the postgres docker image runs the init scripts when it is started up the first time and during that time it won't allow connections.

The Redis errors are valid if the Redis container didn't start up. Now that I thought it through I think if you don't have these docker images down, the first init will be slower. This might be solved with a script that pulls the images first.

I'll investigate some more. Do you think pulling the images could be part of a postinstall script? 🤔

starters/express-typescript-typeorm-postgres/src/controllers/health/handlers/get-health.ts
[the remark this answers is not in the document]
BigAB: Maybe `207`? Whatever WebDAV is 🤷  I don't know.

But I suppose if your health check is requesting multiple things, then the status codes for each of those responses could be included.  

Or I don't know, just 500 anything that isn't 'OK' and have an explanation in the body. 🤷 Yeah not sure.

starters/express-typescript-typeorm-postgres/src/controllers/technology/handlers/create-technology.ts
```diff
@@ -1,20 +1,47 @@
...
+    description: req.body.description,
+  });
+
+  if (inserted.type === Result.ERROR) {
+    LogHelper.error(inserted.message, inserted.error);
+    next(inserted.error);
+    return;
   }
+
+  clearCacheEntry(req.baseUrl);
```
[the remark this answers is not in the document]
BigAB: The cache key would have to include the pagination arguments.  I don't know, cacheing `get-all`s seems pretty normal

starters/express-typescript-typeorm-postgres/src/controllers/technology/handlers/delete-technology.ts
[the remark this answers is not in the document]
BigAB: > I agree on extracting the methods
Just do that.  I get overly excited about over-engineering sometimes. I still think it could be a good change, but we don't have to do it here

starters/express-typescript-typeorm-postgres/src/db/datasource.ts
BigAB: Ah yes I see

starters/express-typescript-typeorm-postgres/src/controllers/technology/handlers/get-technology.ts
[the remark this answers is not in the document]
BigAB: 🤷

starters/express-typescript-typeorm-postgres/docker-compose.yml
BigAB: Personally I do not like postinstall scripts.

starters/express-typescript-typeorm-postgres/src/controllers/technology/handlers/create-technology.ts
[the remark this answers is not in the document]
TapaiBalazs: True, but if something changes you would need to invalidate all the pagination entries which there is no way to know. 🤔

starters/express-typescript-typeorm-postgres/docker-compose.yml
TapaiBalazs: So, I added a postinstall script, but I'm not convinced that I like it either.

I also tried running `dev` with a completely clean setup, with docker images removed and it started up for me. It had a little bit more boot up time, but eventually it got there. Now with a slow internet connection it also pose an issue, that is why I thought of the postinstall script.

But let's discuss it in the afternoon/your morning

starters/express-typescript-typeorm-postgres/src/controllers/health/handlers/get-health.ts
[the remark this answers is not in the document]
TapaiBalazs: I will go with 207 now.

starters/express-typescript-typeorm-postgres/src/controllers/technology/handlers/delete-technology.ts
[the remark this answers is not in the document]
TapaiBalazs: I extracted them, I'll resolve this conversation now

starters/express-typescript-typeorm-postgres/docker-compose.yml
TapaiBalazs: I also tried to "mitigate" the extreme amount of error logs coming from the API not being able to connect to the redis instance. Now it will log it once. I have a log for reconnecting, maybe that could be removed as well, or somehow throttled

starters/express-typescript-typeorm-postgres/src/controllers/health/handlers/get-health.ts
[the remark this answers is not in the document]
TapaiBalazs: So, I asked Dario about his opinion, and he told me that he would prefer to control when the alerting system alerts, therefore, we should return with 200, but in the response body return the error as well.

If the end users of this kit will need an alerting system, they will integrate it, and notify it where they see fit, I added a comment in the code and return with status 200.

starters/express-typescript-typeorm-postgres/src/controllers/technology/handlers/create-technology.ts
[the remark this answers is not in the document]
BigAB: There would have to be some code or convention that "relates" a get-one to a get-list.
But for now this is probably fine

starters/express-typescript-typeorm-postgres/docker-compose.yml
BigAB: SO, whatever it was, running `npm run infrastructure:clear` seemed to fix it for me 🤷 

Maybe it was fine

BigAB (approved):
LGTM

starters/express-typescript-typeorm-postgres/docker-compose.yml
TapaiBalazs: I'm glad it resolved. If you've checked out the branch before when only the database was set up and you created that docker image, that could have caused issues if the docker container wasn't cleared.

Nevertheless, I'll keep an eye out for similar errors. 👀

PR #617, opened 2022-12-07
[Deno] add logging level

### LINKED ISSUE #615: [deno] Add logger
### Acceptance Criteria

- [ ] The app should be using [Deno's logger](https://medium.com/deno-the-complete-reference/using-logger-in-deno-44c5b2372bf3) instead of console.log
- [ ] Some logging settings such as log level should be configurable through an environment variable

tyrelchambers:
## Summary of change

Add log level to Deno logger

closes #615


starters/deno-oak-denodb/src/main.ts
```diff
@@ -1,11 +1,27 @@
-import { Application, Context, oakCors, Router } from '../deps.ts';
+import { Application, Context, oakCors, Router, log, config } from '../deps.ts';
 import { corsAllowedOrigins } from './util/cors_allowed_origins.ts';
 import { API_HOST, PORT } from './config/environment.ts';
 
 const app = new Application();
 const port = +PORT || 3333;
-
+const { LOG_LEVEL } = config({ safe: true });
```
vyktoremario, line 7: This can be exported from `./config/environment.ts` and added to the import in line 3

starters/deno-oak-denodb/src/main.ts
```diff
@@ -1,11 +1,27 @@
...
 const router = new Router();
+
+await log.setup({
```
vyktoremario, line 10: Is it possible to move this log setup to its own module?

vyktoremario (changes requested):
Left some comments

starters/deno-oak-denodb/.env.example
```diff
@@ -9,3 +9,4 @@ DATABASE_PORT=5432
 
 
 CORS_ALLOWED_ORIGINS=http://localhost:3333
+LOG_LEVEL="DEBUG"
```
dariodjuric, line 12: Note that you don't need to use quotes here - the parser will work either way.

dariodjuric (changes requested):
Do you think we could also replace the usages of `console.log` with this logger?

dariodjuric (approved):

PR #625, opened 2022-12-08
[NODE] feat: initial setup of bullMQ

### LINKED ISSUE #618: [node-pod] Add Queuing
# Description
(these requirements come from meeting with Dustin 12/7/22) We will have a POST endpoint which will accept something that will add the job to BullMQ with the package name. With BullMQ we will automatically listen for changes in the database and run the jobs in FIFO ordering. Jobs should run in worker threads.

Note: there is a possibility that Redis client does not work well with BullMQ, but this will be determined in further research. 

# Notes
General notes from specifications:

<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-41132257-7fff-aff7-9b4f-46315d612cd7"><p dir="ltr" style="line-height:1.38;margin-top:0pt;margin-bottom:0pt;"><span style="font-size:11pt;font-family:Arial;color:#000000;background-color:transparent;font-weight:400;font-style:normal;font-variant:normal;text-decoration:none;vertical-align:baseline;white-space:pre;white-space:pre-wrap;">As applications grow, the need for asynchronous workers become needed. Newer developers are less familiar with this pattern. As such, we want to include this as part of the starter kit as an opt-in feature so it can be easily used by developers.</span></p><br /><div dir="ltr" style="margin-left:0pt;" align="left">

Requirement | Description
-- | --
Select Queue Tech | There are several queue solutions on the market that serve different purposes. Select a queue that makes the most sense for the project and add it.
Infrastructure Setup | Regardless of the queue selected, a message queue will need to be utilized. For push queues, this is typically done with a formal message queue tech such as ElasticMQ, RabbitMQ, etc. For pull queues, this is typically done with Redis or equivalent in-memory storage. All of these require some level of local setup. Add this setup to the docker compose with setup instructions.
Add job generator endpoint | Add a POST /example-job endpoint that receives a message and adds it as a job to the queue to be processed.
Add worker/job processor | Add a worker that can process the job added to the queue. It should simply print the job message to STDOUT.
Add queue runner | Add or modify scripts to run the queue.

</div></b>

TapaiBalazs:
## Type of change

- [ ] Documentation change
- [ ] Bug fix
- [X] Feature

## Summary of change

- Added another redis instance
- set up bullMQ connection
- set up basic queue controller
- set up worker to run on different thread

## Checklist

- [x] This fix resolves #618 

### DIFF
--- a/starters/express-typescript-typeorm-postgres/src/controllers/health/handlers/get-health.ts
+++ b/starters/express-typescript-typeorm-postgres/src/controllers/health/handlers/get-health.ts
@@ -4,12 +4,14 @@ import { redisHealthCheck } from '../../../cache/redis-cache-client';
 import { Result } from '../../../constants/result';
 import { dataSource } from '../../../db/datasource';
 import { ErrorResult, SuccessResult } from '../../../interfaces/results';
+import { defaultQueue } from '../../../queue/queue';
 import { LogHelper } from '../../../utils/log-helper';
 
 export async function getHealth(req: Request, res: Response): Promise<void> {
-  const [databaseVersion, redisPingResult] = await Promise.all([
+  const [databaseVersion, redisPingResult, jobCount] = await Promise.all([
     checkDatabaseConnection(),
     checkRedisCacheConnection(),
+    defaultQueue.count(),
   ]);
 
   if (databaseVersion.type === Result.ERROR) {
@@ -39,6 +41,7 @@ export async function getHealth(req: Request, res: Response): Promise<void> {
   res.json({
     database: databaseVersion.data,
     redisConnection: redisPingResult.data,
+    jobCount,
   });
 }
 
--- a/starters/express-typescript-typeorm-postgres/src/controllers/health/handlers/get-health.ts
+++ b/starters/express-typescript-typeorm-postgres/src/controllers/health/handlers/get-health.ts
@@ -1,17 +1,17 @@
 import { Request, Response } from 'express';
 import { getReasonPhrase, StatusCodes } from 'http-status-codes';
-import { redisHealthCheck } from '../../../cache/redis-cache-client';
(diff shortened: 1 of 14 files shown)


starters/express-typescript-typeorm-postgres/src/controllers/health/handlers/get-health.ts
```diff
@@ -20,7 +22,15 @@ export async function getHealth(req: Request, res: Response): Promise<void> {
     return;
   }
 
-  if (redisPingResult.type === Result.ERROR) {
+  if (redisQueueHealthResult.type === Result.ERROR) {
+    res.status(StatusCodes.SERVICE_UNAVAILABLE).json({
```
dariodjuric, line 26: Tip: you can just return this line and that way you don't need to have separate lines for setting the response and returning.

starters/express-typescript-typeorm-postgres/src/controllers/health/services/health.service.ts
```diff
@@ -0,0 +1,95 @@
...
+        type: Result.ERROR,
+        message: error.message,
+        error,
+      };
+    });
+}
+
+function pingRedisQueueWithTimeout(): Promise<'PONG'> {
+  return defaultQueue.client.then((client: RedisClient) => {
+    return Promise.race([
```
dariodjuric, line 84: Nice trick.

dariodjuric (approved):
LGTM

starters/express-typescript-typeorm-postgres/src/controllers/health/handlers/get-health.ts
TapaiBalazs: Yes, that's true, I'll do it in a separate PR though with all the handlers

PR #627, opened 2022-12-09
Fix/621 unique docker containers

### LINKED ISSUE #621: [node-pod] Docker container uniqueness setup
# Updated requirements based on discussions with @dustinsgoodman 

1. Cleaning up the containers should not be part of the starter kit as scripts
2. Document in the readme how they can clean up and completely reset databases/caches, etc
3. Remove named containers from the docker-compose.yml file
4. Separate infrastructure instantiation from the `dev` script

This way makes it explicit to the user that there are infrastructure needs to run the server and gives them a place to explore.
Removing the container names will solve the uniqueness constraints, because the container name should be inferred from the CWD of the docker-compose.yml file.

##  Original ticket contents just for records:
The express-typeorm-postgres starter sets up a redis and a postgres docker container for the project. These containers are named containers, because we have clear scripts, for example:

- npm run cache:delete - deletes the cache data (mounted folder to the docker container), and the cache image itself
- npm run db:delete - deletes the cache data (mounted folder to the docker container), and the database image itself
- npm run infrastructure:clear - deletes everything: mounted folders, docker containers, built images.

The above scripts (well the infrastructure one uses the others) need to know the docker container's name to effectively clean up, that is why they are set in the docker-compose.yml file as follows:

```
  postgres:
    image: 'postgres'
    container_name: 'data_storage'

# ...

  redis_cache:
    image: 'redis:alpine'
    container_name: 'cache'
```

This setup would cause problems for users who create more than one projects with this kit. Whichever project starts up first, the other one will use the same containers and there are going to be issues, because the two projects probably will have different database schemas and tables.

Therefore, the generator script needs to be updated to do the following tasks, if the selected kit is the express-typeorm-postgres:

1. Read the docker-compose.yml and replace every `container_name: 'cache'` lines to `container_name: 'projectname_cache`. (Same with the queue redis image name and the database image name)
5. Iterate over the scripts of the package json, and replace every `docker rm cache -f` with `docker rm projectname_cache -f` (Same with the queue redis image name and the database image name)

TapaiBalazs:
## Type of change

- [X] Documentation change
- [ ] Bug fix
- [X] Feature

## Summary of change

- Applied changes for starting up dev mode and infrastructure
- Changed docker-compose to not include container names
- Set up misc folder for infrastructure related data
- Moved entities folder under the db folder
- Updated readme

## Checklist

- [ ] This fix resolves #621  #583 


starters/express-typescript-typeorm-postgres/.env.example
```diff
@@ -2,22 +2,34 @@
 
 NODE_ENV=development
 
+# Provide a comma separated list of log levels to control what levels log should appear.
```
dariodjuric, line 5: Nice comments.

Issue #624, opened 2022-12-08
[deno] Add cache


dariodjuric:
We should add a caching layer that stands between the API and the database.

Read the "Data Store" chapter here: https://docs.google.com/document/d/1iEXK6BFgu5HZ4dfU7pfv80zLOAd86T8AzstCMEYTUXA/edit#heading=h.8p25jh2al4b9

### Acceptance Criteria

- [ ] Decide on the caching database
- [ ] Add a local caching database alongside Postgres
- [ ] Create a service for reading/writing to cache
- [ ] In the technology update API, write the updated technology to cache
- [ ] In the technology query API, read the updated technology from cache, otherwise read from DB

PR #630, opened 2022-12-09
[Deno] Add Cache

vyktoremario:
## Type of change

- [ ] Documentation change
- [ ] Bug fix
- [x] feature

## Summary of change

closes: #624

## Checklist

- [ ] This fix resolves #


starters/deno-oak-denodb/src/cache/Cache.ts
```diff
@@ -0,0 +1,157 @@
...
+		typeDefs: any;
+		resolvers: any;
+	};
+	redisClient: Redis;
+	usePlayground: boolean;
+	route: string;
+	defaultEx?: number | undefined;
+};
+
+export class Cache {
```
dariodjuric, line 28: It seems like this class is doing a bit too much. By the name of it, it should only be a wrapper around the Redis library, but here it's also handling GraphQL requests.

I would suggest to have this class perform the functionality of the "Cache API Wrapper" from the [specification](https://docs.google.com/document/d/1iEXK6BFgu5HZ4dfU7pfv80zLOAd86T8AzstCMEYTUXA/edit#), where its public interface only has the following methods from the spec:

- Read item
- Write item
- Invalidate item

Also, the constructor can probably be without arguments. It can instantiate a new Redis instance internally.

starters/deno-oak-denodb/src/cache/Cache.ts
```diff
@@ -0,0 +1,157 @@
+import {
```
dariodjuric, line 1: Shouldn't the file be named `cache.ts` to be consistent with the Deno best practices guide?

dariodjuric (changes requested):
I left some suggestions. Also, `deno lint` is failing.

starters/deno-oak-denodb/docker-compose.yml
```diff
@@ -12,6 +12,21 @@ services:
...
+  redis:
+    image: 'redis:alpine'
+
+    command: redis-server
+
+    ports:
+      - '6379:6379'
+
+    volumes:
+      - $PWD/redis-data:/var/lib/redis
```
TapaiBalazs, line 25: I've read that `$PWD` won't work on a windows machine, I'd suggest using `./redis-data:/var/lib/redis` instead.

starters/deno-oak-denodb/src/cache/cache.ts
```diff
@@ -0,0 +1,63 @@
...
+		return null;
+	}
+
+	async writeItem<T>(
+		{ cacheKey, ex }: { cacheKey: string; ex?: number },
+		callback: GetFreshValue<T>,
+	): Promise<T> {
+		const results = await callback();
+		let redisCachingOpts: SetOpts = { keepttl: true };
+		console.log('results', results);
```
TapaiBalazs, line 40: Do we need this console.log here?

starters/deno-oak-denodb/src/cache/cache.ts
```diff
@@ -0,0 +1,63 @@
...
+
+		// Set the expiration time for the individual request if provided
+		if (ex) {
+			if (ex > 0) redisCachingOpts = { ...redisCachingOpts, ex };
+		}
```
TapaiBalazs, line 45: Correct me if I'm wrong, but if `ex` is `0` then `if (ex)` will be falsy, therefore won't go into the block. This also means that this can be simplified?

```
if (ex) {
  redisCachingOpts = { ...redisCachingOpts, ex }
}
```

dariodjuric (changes requested):
I left some more comments.

starters/deno-oak-denodb/src/cache/cache.ts
```diff
@@ -0,0 +1,63 @@
...
+
+type GetFreshValue<T> = {
+	(): Promise<T> | T;
+};
+
+const redisHostname = REDIS_CACHE_HOST || 'localhost';
+const redisPort = REDIS_CACHE_PORT ? parseInt(REDIS_CACHE_PORT) : 6379;
+
+export class Cache {
+	redisClient!: Redis;
```
TapaiBalazs, line 13: I'd expose the `ping` method in this class as well if there's one in the deno redis implementation, because that's useful for the health endpoint. WDYT?

starters/deno-oak-denodb/src/cache/cache.ts
```diff
@@ -0,0 +1,63 @@
...
+		const cacheValue = await this.redisClient.get(cacheKey);
+		if (cacheValue) {
+			return JSON.parse(cacheValue);
+		}
+		return null;
+	}
+
+	async writeItem<T>(
+		{ cacheKey, ex }: { cacheKey: string; ex?: number },
+		callback: GetFreshValue<T>,
```
dariodjuric, line 36: Why do we have a callback here instead of passing just the regular value to be cached?

starters/deno-oak-denodb/src/cache/cache.ts
dariodjuric, line 40: It looks like this one was put here for debugging purposes.

starters/deno-oak-denodb/src/graphql/resolvers/mutation_handler.ts
```diff
@@ -15,17 +18,26 @@ export const createTechnology = async (
...
 	_: unknown,
 	{ id, input }: { id: string; input: TechnologyInput },
-) => {
+	{ ds }: { ds: Cache },
+	info: Record<string, string>,
+): Promise<{ done: boolean }> => {
 	await Technologies.where('id', id).update({
 		...input,
 	});
+	await ds.invalidateItem(`${info.fieldName}:${id}`);
```
TapaiBalazs, line 27: When updating an entry, should you invalidate the `getTechnologies` entry as well? 🤔

starters/deno-oak-denodb/src/graphql/resolvers/mutation_handler.ts
```diff
@@ -15,17 +18,26 @@ export const createTechnology = async (
...
 
-export const deleteTechnologyById = async (_: unknown, { id }: { id: string }) => {
+export const deleteTechnologyById = async (
+	_: unknown,
+	{ id }: { id: string },
+	{ ds }: { ds: Cache },
+	info: Record<string, string>,
+): Promise<{ done: boolean }> => {
 	await Technologies.deleteById(id);
+	await ds.invalidateItem(`${info.fieldName}:${id}`);
```
TapaiBalazs, line 40: When we delete an item, should we invalidate the `getTechnologies` entry as well?

starters/deno-oak-denodb/src/graphql/resolvers/query_handler.ts
```diff
@@ -1,11 +1,34 @@
+import { Cache } from '../../cache/cache.ts';
 import { Technologies } from '../../db/model/technology.ts';
 
-export const getTechnologies = async () => {
-	const technologies = await Technologies.all();
-	return technologies;
+export const getTechnologies = async (
+	_p: unknown,
```
TapaiBalazs, line 5: Is this a deno specific syntax for using `_p` and `_` as parameter names? 🤔  What's the reason for not having something descriptive?

starters/deno-oak-denodb/src/graphql/resolvers/query_handler.ts
```diff
@@ -1,11 +1,34 @@
...
+	_: unknown,
+	{ ds }: { ds: Cache },
+	info: Record<string, string>,
+): Promise<Technologies[]> => {
+	const cacheTechnologies = await ds.readItem<Technologies[]>({ cacheKey: info.fieldName });
+	if (cacheTechnologies) {
+		return cacheTechnologies;
+	}
```
TapaiBalazs, line 13: This "read item from cache an check if it is cached" routine is very repetitive. I know it is now only two endpoints, but a generic higher-order function might be good to have. WDYT?

starters/deno-oak-denodb/src/graphql/resolvers/query_handler.ts
```diff
@@ -1,11 +1,34 @@
...
+	_: unknown,
+	{ id }: { id: string },
+	{ ds }: { ds: Cache },
+	info: Record<string, string>,
+): Promise<Technologies> => {
+	const cacheTechnology = await ds.readItem<Technologies>({ cacheKey: `${info.fieldName}:${id}` });
+	if (cacheTechnology) {
+		return cacheTechnology;
+	}
+	return ds.writeItem({ cacheKey: `${info.fieldName}:${id}` }, async () => {
```
TapaiBalazs, line 30: see my previous comment on repeating code

TapaiBalazs (commented):
Looks good, I left some comments for consideration.

starters/deno-oak-denodb/src/graphql/resolvers/query_handler.ts
vyktoremario: Adding an underscore in front of the variable name tells deno lint that it is an unused variable that is needed due to positional argument

dariodjuric (approved):

PR #638, opened 2022-12-12
[NODE] Refactor/572 rename starter kit

### LINKED ISSUE #572: [node-pod] Rename folder and starter kit to express-typeorm-postgres
As per Dustin's request, the folder and everything related to the starter kit's name should be renamed

from `express-typescript-typeorm-postgres` to `express-typeorm-postgres`

Additional tasks: 

- [ ] Add a .nvmrc file with node 16 latest to the folder

TapaiBalazs:
## Type of change

- [X] Documentation change
- [ ] Bug fix
- [X] Refactor

## Summary of change

Renamed the folder that contains this kit to not include typescript in the name and to follow the framework-schema-database structure as in the specification.

## Checklist

- [x] This fix resolves #572 


starters/express-typeorm-postgres/package.json
```diff
@@ -1,11 +1,10 @@
 {
-  "name": "express-typescript-typeorm-postgres",
+  "name": "express-typeorm-postgres",
   "version": "0.0.1",
   "description": "express, typescript, REST",
   "author": "This Dot Labs",
   "license": "MIT",
   "scripts": {
-    "postinstall": "docker-compose pull",
```
tvanantwerp, line 8: Curious about this—not sure what it was for, so equally not sure why it needs to go.

tvanantwerp (approved):
Very straightforward changes, so 👍.

Curious about that `postinstall` script removal, as the PR doesn't mention what that's about. But I don't see them in use anywhere else in starter.dev, so I'm not very concerned.

vyktoremario (approved):

starters/express-typeorm-postgres/package.json
TapaiBalazs: Yes, It is a remnant of a PR before this, that I forgot to remove before and I noticed it now. Basically, if the user does not have docker installed or running, this would fail their npm installs, and since in the previous PR we removed the "need" for pulling docker images by splitting up infrastructure and development mode API serving into two different commands we can remove this liability.

PR #640, opened 2022-12-13
[Deno] Add healthcheck

### LINKED ISSUE #567: [deno] 3. Create healthcheck
# Background

All APIs should come with a default healthcheck endpoint. The healthcheck should provide a response indicating that all systems are operational. If a data source is involved, respond with data from the data source or something about that connection.

# Acceptance

- [ ] There should be an endpoint or graphql query that provides a healthcheck response

dariodjuric:
Resolves #567.


starters/deno-oak-denodb/src/rest/services/healtcheck_service.ts
```diff
@@ -0,0 +1,39 @@
...
+	let isRunning;
+
+	try {
+		isRunning = await cache.pingRedis() === 'PONG';
+	} catch (e) {
+		isRunning = false;
+		logger.error(e);
+	}
+
+	return isRunning;
```
TapaiBalazs, line 16: So far, so good.

Have you checked what happens when you stop the redis container?
In node world with the node redis client the healthcheck endpoint was spinning because it was waiting for the request to time out. If the deno redis client works the same, this might be improved with a Promise.race trick.

WDYT? 🤔

starters/deno-oak-denodb/src/rest/services/healtcheck_service.ts
```diff
@@ -0,0 +1,39 @@
...
+	let isRunning;
+
+	try {
+		const connector = db.getConnector() as PostgresConnector;
+		// We're accessing a private property here, but this is to work around a bug where the connector ping()
+		// method is not implemented properly (below is the proper implementation)
+		// This should be fixed in this PR: https://github.com/eveningkid/denodb/pull/363
+		const client = connector._client;
+		const [firstRow] = (
+			await client.queryObject<{ result: number }>('SELECT 1 + 1 as result')
```
TapaiBalazs, line 29: I went with `SELECT version()` so in the health check it provides information about the database in the node kit. This seems to be okay to me as well, just a note.

TapaiBalazs (approved):
LGTM, I added some comments on what I did differently in the node kit, just for consideration.

vyktoremario (approved):

starters/deno-oak-denodb/src/rest/services/healtcheck_service.ts
dariodjuric: Yeah, good idea - I added that.

PR #655, opened 2022-12-14
Add Deno/Oak/DenoDB starter kit

dariodjuric:
## Type of change

- [ ] Documentation change
- [ ] Bug fix
- [x] New starter kit

## Summary of change

This adds a new starter kit for Deno, and makes some adjustments in the CLI to support non-Node starter kits.

## Checklist

- [ ] This fix resolves #


dustinsgoodman:
# December 14, 2022 Review

* I really enjoyed how the express kit by @TapaiBalazs documented the .env file. I would like that here too.
* If you pass the `-d` flag to `docker compose up`, it'll run the process in the background and not lock up the user's terminal. I recommend doing this instead.
* Running the docs command I only get 1 file. Is this intentional?
image
* I'd like to see default routes in the README so a user knows how to interact quickly with this codebase.
* The GraphQL implementation fights best practice conventions. `getTechnologies` for instance should just be `technologies`. Ditto for the singular query. Mutations are fine.
* Did we manage to do a queue?
* The cache doesn't seem to work out of the box: 
Screenshot 2022-12-14 at 3 28 01 PM
* Notes on the database and extending it through migrations and seeding would be appreciated.

packages/create-starter/src/index.ts
```diff
@@ -74,30 +74,43 @@ export async function main() {
...
-    EXCLUDED_PACKAGE_JSON_FIELDS.forEach((field) => delete packageJSON[field]);
-
-    try {
-      await fs.writeFile(path.join(destPath, 'package.json'), JSON.stringify(packageJSON, null, 2));
-      await overrideAngularJsonIfExists(destPath, options.kit, options.name);
-
-      await initGitRepo(destPath);
-    } catch (_) {
-      console.info(gray(`> ${bold('Note:')} Failed to update package.json. You may need to do this manually.`));
+    const packageJsonPath = path.join(destPath, 'package.json');
```
dariodjuric, line 77: @danecando, this is what we talked about a few days ago. I made this adjustment for non-Node starter kits. Does this look OK to you?

dariodjuric:
> * I really enjoyed how the express kit by @TapaiBalazs documented the .env file. I would like that here too.

Done!

> * If you pass the `-d` flag to `docker compose up`, it'll run the process in the background and not lock up the user's terminal. I recommend doing this instead.

Done! The reason it was done this way was to have the `start-all` command to shut down all the containers on Ctrl+C, but this also works. 

> * Running the docs command I only get 1 file. Is this intentional?

Yes, this was just a showcase of how to generate documentation. But we took this opportunity to document the API handlers as well, and note in the readme that the documentation is based on JSDoc comments.

> * I'd like to see default routes in the README so a user knows how to interact quickly with this codebase.

Good idea, done!

> * The GraphQL implementation fights best practice conventions. `getTechnologies` for instance should just be `technologies`. Ditto for the singular query. Mutations are fine.

Good point, done!

> * Did we manage to do a queue?

Sadly, no. The Deno ecosystem does not have a robust queuing library like Node does, so we've decided to leave it out for now. We even had to roll out our own caching wrapper based on Redis.

> * The cache doesn't seem to work out of the box:

Great catch! Fixed.

> * Notes on the database and extending it through migrations and seeding would be appreciated.

Done! DenoDB doesn't support migrations, but we've documented how to seed and add new models in more detail.

> * When I hit save on the deno.json, it reformats.

Originally it was using spaces while everything else was using tabs. Fixed by changing to tabs, great catch!

dustinsgoodman (approved):
Just some responses to the feedback but overall very pleased with this and I'm giving my stamp of approval!

> > If you pass the -d flag to docker compose up, it'll run the process in the background and not lock up the user's terminal. I recommend doing this instead.
>
> Done! The reason it was done this way was to have the start-all command to shut down all the containers on Ctrl+C, but this also works.

I actually don't want the start-all script at the end of the day. Having to wait for infra to spin up and down is unnecessary and makes the DX not as good.


> > Did we manage to do a queue?
> 
> Sadly, no. The Deno ecosystem does not have a robust queuing library like Node does, so we've decided to leave it out for now. We even had to roll out our own caching wrapper based on Redis.

All good. Let's add a ticket for a one day thing when they solve this and move on. I can accept this but something we should talk to them about. 😄 

> > Notes on the database and extending it through migrations and seeding would be appreciated.
> 
> Done! DenoDB doesn't support migrations, but we've documented how to seed and add new models in more detail.

Technically speaking - sync is a migration! So that's what I wanted. Great! 🎉

starters/deno-oak-denodb/README.md
```diff
@@ -0,0 +1,322 @@
...
+deno task generate-type-definition
+```
+
+## Using the GraphQL API
+
+The GraphQL API playground is available at <http://localhost:3333/graphql>. You can execute the following query to retrieve all technologies, assuming that they were seeded properly.
+
+```graphql
+{
+  getTechnologies {
```
vyktoremario, line 248:
```suggestion
  technologies {
```

PR #706, opened 2022-12-21
chore(website): Astro Upgrade

danecando:
## Changes

- Upgrades to latest version of Astro (v1) from Astro v1-beta

https://docs.astro.build/en/migrate/#astro-10

Closes https://github.com/thisdot/pods/issues/125

### DIFF
--- a/packages/website/.gitignore
+++ b/packages/website/.gitignore
@@ -18,4 +18,4 @@ yarn-error.log*
 .DS_Store
 
 # ignore generated kit pages
-src/pages/kits/*.md
+src/pages/kits/*.mdx
--- a/packages/website/.gitignore
+++ b/packages/website/.gitignore
@@ -18,4 +18,5 @@ yarn-error.log*
 .DS_Store
 
 # ignore generated kit pages
+src/pages/kits/*.md
 src/pages/kits/*.mdx
--- a/packages/website/scripts/generate-kit-pages.js
+++ b/packages/website/scripts/generate-kit-pages.js
@@ -50,7 +50,7 @@ ${markdown}`;
   try {
     const handle = await fs.opendir(kitPagesPath);
     console.info('kitgen: deleting existing kit pages');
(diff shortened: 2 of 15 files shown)


packages/website/.gitignore
```diff
@@ -18,4 +18,4 @@ yarn-error.log*
 .DS_Store
 
 # ignore generated kit pages
-src/pages/kits/*.md
+src/pages/kits/*.mdx
```
TapaiBalazs, line 21: Is there anything else that needs _not_ to be ignored in that folder? 🤔 

If previously these files were md files, and somebody in their local branches pull down the new changes, the lingering md files from the previous version can be accidentally committed. Therefore, I'd suggest adding the .mdx rule and keeping the .md as well, or just ignore the whole folder content if everything there should be ignored.

What do you think?

packages/website/scripts/generate-kit-pages.js
```diff
@@ -50,7 +50,7 @@ ${markdown}`;
   try {
     const handle = await fs.opendir(kitPagesPath);
     console.info('kitgen: deleting existing kit pages');
-    rimraf(`${kitPagesPath}/*.md`, async (err) => {
+    rimraf(`${kitPagesPath}/*.md,${kitPagesPath}/*.mdx`, async (err) => {
```
TapaiBalazs, line 53: Is it worth leaving the `.md` files here? 🤔 It would clean up old remnant files in one's dev environment, that's true.

TapaiBalazs (approved):
LGTM, I left some minor comments/questions but nothing show stopper.

packages/website/scripts/generate-kit-pages.js
danecando: Couldn't hurt. Then people don't have to worry about deleting the old files themselves.

TapaiBalazs (approved):
👍

PR #715, opened 2023-01-04
decrement shows negative values and also a test to do same implemented

hdJerry:
# Description
- currently, the decrement button does not allow numbers to go past zero into negative numbers. The app should allow going into the negative


Megio (approved):
LGTM

SDaian (approved):
LGTM

PR #801, opened 2023-01-17
Qwik/greeting component issue fix

hdJerry:
Image

# Description
Greeting component has an input field and doesn't accept the greeting query parameter
- it should not accept an input field
- it should be dynamic to query param as shown in the screenshot

## Result

Screenshot 2023-01-17 at 11 43 58
Screenshot 2023-01-17 at 11 44 21


SDaian (approved):
LGTM 👍

Issue #815, opened 2023-01-18
[solidjs-tailwind] Lint all files in starter kit


kenslachtajr:
# Description
- kit should be free of linting errors and all files should have been linted. Currently, 2 files need to be linted to change the formatting to match what the linter is configured to suggest. You see these when you run `yarn format`

Issue #814, opened 2023-01-18
[solidjs-tailwind] Fix linting error


kenslachtajr:
# Description
- all linting should pass without errors. Currently there is a lint warning when running `yarn lint`. 

![Image](https://user-images.githubusercontent.com/1828602/213083206-302071db-93c4-4e93-a1b4-9b0135da4c8f.png)

PR #828, opened 2023-01-19
solidjs-fix: linting warning

Megio:
## Type of change

Fix linting warning and files.

Fixes: #815
Fixes: #814


hdJerry (changes requested):
I know this is not part of the ticket but please can you update this part of the code as its not working as expected

```jsx
<input
          type="text"
          value={greeting()}
          placeholder="Who should the server greet?"
          onKeyUp={(ev) => {
            setTimeout(() => setGreeting(ev.target.value), 1000)
          }}
          class="text-l w-96 max-w-full rounded border-0 bg-white px-4 py-4 shadow outline-none placeholder:italic focus:outline-none focus:ring"
        />
```

Just take note of the difference.

Thanks

hdJerry (approved):
I noticed there is a ticket to remove that input completely.

Ignore previous message. Thanks

PR #836, opened 2023-01-20
[serverless-apollo-contentful] Add missing tests for the kit

### LINKED ISSUE #826: [serverless-apollo-contentful] Add missing tests for the kit
- [ ] Add missing tests for src/schema/post and src/schema/comment
- [ ] Update README.md

Image

Image

Image

kodejuice:
## Type of change

Add tests

Closes #826

- [ ] Documentation change
- [ ] Bug fix

## Checklist

- [x] This fix resolves #826


starters/graphql-serverless-contentful/.jest/set-env-vars.ts
```diff
@@ -0,0 +1,6 @@
+import * as dotenv from 'dotenv';
+
+// load dotenv config
+const { parsed } = dotenv.config();
+
+process.env = { ...process.env, ...parsed };
```
mbicknese, line 6: There's no newline at the end of this file. Do you have your editor/IDE configured to use the `.editorconfig` file?

starters/graphql-serverless-contentful/src/schema/post/post.resolver.test.ts
```diff
@@ -0,0 +1,209 @@
...
+			subject = await apolloServer.executeOperation({
+				query: QUERY,
+				variables: {
+					id: post.id,
+				},
+			});
+		});
+
+		it('returns the post with the given id', () => {
+			// console.log('by id', JSON.stringify(subject));
```
mbicknese, line 142: Leftover console statement.

mbicknese (changes requested):
Can we discuss this PR in voice chat? I'd like to know more about the setup you've created.

starters/graphql-serverless-contentful/.jest/set-env-vars.ts
mbicknese: Also have a look at @ihardz comment here, that's a good change as I don't expect devs to have set up their environment or dot files for the tests to work.

https://github.com/thisdot/starter.dev/pull/835/files#r1082622089

starters/graphql-serverless-contentful/src/schema/post/post.resolver.test.ts
```diff
@@ -0,0 +1,204 @@
...
+				content,
+			},
+		});
+
+		assert(res.body.kind === 'single');
+		expect(res.body.singleResult.errors).toBeUndefined();
+		assert(res.body.singleResult.data);
+
+		post = res.body.singleResult.data.createPost;
+	});
```
mbicknese, line 89: I might have only discussed this with @iansamz before you joined the call. The mutation is not required for the other tests to succeed. Seeing we have the persistence of data completely mocked, it wouldn't store the result of this mutation anyway. We can safely wrap it in its own test without it having to be `beforeAll`.

All other tests would continue to work, as the mock contentful will return the default posts and comment. We just need to refer to the mocked values directly instead of having the `post` references.

This benefits us because one no longer needs to figure out why this specific test needs to run before all other tests. 

EDIT: I just found that we already have this test as an actual test further down the line. Feel free to completely remove the `beforeAll` block.

starters/graphql-serverless-contentful/src/schema/post/post.resolver.test.ts
```diff
@@ -0,0 +1,204 @@
...
+						id: post.id,
+						content: post.content,
+					},
+				],
+			});
+		});
+	});
+
+	describe('mutation createPost', () => {
+		let created_post: any;
```
mbicknese, line 149: This definition can move to line 170 and be turned into a `const`
But also see my other comment of asserting `subject.body.singleResult.data.createPost` directly

starters/graphql-serverless-contentful/src/schema/post/post.resolver.test.ts
```diff
@@ -0,0 +1,204 @@
...
+import { apolloServer } from '../../handlers/graphql';
+import { GraphQLResponse } from '@apollo/server/src/externalTypes';
+import assert from 'assert';
+
+const MOCK_CONTENT = [
+	'This is the test post content',
+	'This is the updated test post content',
+];
+
+jest.mock('../../utils/contentful', () => {
```
mbicknese, line 11: This comment is a nice to have. Feel free to leave things as is.
I just remembered how we could turn this around, so we don't need the double declaration anymore. Here's a small example. If that doesn't clarify things, just ping me.

```JavaScript
import { myService } from '../myService';

jest.mock('../myService', () => ({ getClient: jest.fn() }));
const MOCKED_COMMENTS = ['This is a cool comment', 'This is an angry comment'];
myService.getClient.mockReturnValue({ getEntries: () => MOCKED_COMMENTS });
```
```

mbicknese: cc @iansamz, you might like this as well.

Issue #884, opened 2023-01-26
[solidjs-tailwind] Remove package specific scripts from README


kenslachtajr:
# Description
- all scripts should be package agnostic. Remove `pnpm` from the README

![Image](https://user-images.githubusercontent.com/1828602/214954476-b4bcc37b-3665-49ef-8703-516a511ebeef.png)

Issue #883, opened 2023-01-26
[solidjs-tailwind] Put "Repository" and "Live demo on separate lines in README


kenslachtajr:
# Description
- put the two links on separate lines, otherwise it's confusing

![Image](https://user-images.githubusercontent.com/1828602/214949005-2e7a76c0-f61f-44cf-bde9-a44370bdbab7.png)

Issue #882, opened 2023-01-26
[solidjs-tailwind] Fix Storybook script typo in README


kenslachtajr:
# Description
- `pnpm run storbook` - To showcase the component library - typo in README

Issue #881, opened 2023-01-26
[solidjs-tailwind] Reword overview section of README


kenslachtajr:
# Description
- need to reword first part of README overview as follows:

> "This starter kit features two of the hottest frameworks in the JavaScript and CSS ecosystems. SolidJS is a new framework built on top of the React API and offers a leaner and more performant alternative to React."

PR #889, opened 2023-01-27
solidjs-fix: readme

Megio:
Fix readme file.

Fixes: #884 
Fixes: #883 
Fixes: #882 
Fixes: #881

### DIFF
--- a/starters/solidjs-tailwind/README.md
+++ b/starters/solidjs-tailwind/README.md
@@ -37,14 +37,15 @@ This starter kit features two of the hottest frameworks in the JavaScript and CS
 
 ### CLI (Recommended)
 
+Run these commands with your favorite package manager
+
 ```bash
-npm i -g pnpm
-pnpm create @this-dot/starter --kit solidjs-tailwind
+create @this-dot/starter --kit solidjs-tailwind
 ```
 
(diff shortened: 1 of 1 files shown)


hdJerry (changes requested):
There should be information that tells the user to use either npm, yarn or pnpm. We can't just take out pnpm run from the command. 😃

Megio:
> There should be information that tells the user to use either npm, yarn or pnpm. We can't just take out pnpm run from the command. 😃

What if I add under the the `Commands` title something like "Run this commands with your favourite package manager" and then the list of pure commands?

markshenouda (approved):

starters/solidjs-tailwind/README.md
```diff
@@ -35,6 +35,8 @@ This starter kit features the most two hot frameworks in JavaScript and CSS echo
 
 ## Installation
 
+This README is written with pnpm in mind but you can use other manager runner
```
hdJerry, line 38: Please add the possible package manager. And put you can give an example too. Like npm run dev, or pnpm run dev or yarn dev

vyktoremario (approved):

hdJerry (approved):

PR #913, opened 2023-01-30
[express-apollo-prisma] Final review and deployment

### LINKED ISSUE #867: [express-apollo-prisma] Final review and deployment
# Description
- this is the final review and deployment ticket for the kit

ihardz:
## Type of change

- [ ] Documentation change
- [ ] Bug fix

## Checklist

- [ ] This fix resolves #867

### DIFF
--- a/starters/express-apollo-prisma/README.md
+++ b/starters/express-apollo-prisma/README.md
@@ -216,17 +216,20 @@ type Mutation {
 `;
 ```
 
-   You can download the [Apollo GraphQL extension](https://marketplace.visualstudio.com/items?itemName=apollographql.vscode-apollo) that adds syntax highlighting for GraphQL files and gql templates inside JavaScript files.
+You can download the [Apollo GraphQL extension](https://marketplace.visualstudio.com/items?itemName=apollographql.vscode-apollo) that adds syntax highlighting for GraphQL files and gql templates inside JavaScript files.
 
 4. Update our automatically generated code by running `npm run codegen`. This will generate a new `src/graphql/schema/index.ts` file.
+
 ```sh
 npm run codegen
 ```
+
 5. Finally, update your resolvers with updated properties you'd like to be returned by GraphQL.
 
 ### Pagination
 
-There is a `technologies` query in `technology.typedefs.ts` file. The query uses offset pagination for querying results. It is a place for configuring the pagination defaults. Current defaults: 
+There is a `technologies` query in `technology.typedefs.ts` file. The query uses offset pagination for querying results. It is a place for configuring the pagination defaults. Current defaults:
+
 - limit: 5
 - offset: 0
 
@@ -263,7 +266,7 @@ In order to restrict origin URLs that can access your API, you need to add a lis
     - /data-sources/** - holds data sources per model.
     - /mappers/** - holds mappers per model.
     - /schema/** - holds GraphQL schema type definitions and resolvers.
-    - /server-context/** - holds server-context types and middleware factory methods.    
+    - /server-context/** - holds server-context types and middleware factory methods.
   - /redis/** - holds Redis-related functionality.
   - main.ts - bootstraps an entire server.
 ```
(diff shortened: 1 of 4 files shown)


starters/express-apollo-prisma/src/graphql/data-sources/technology-data-source.ts
```diff
@@ -0,0 +1,97 @@
...
+		id: TechnologyEntityId,
+		updateTechnology: Prisma.TechnologyEntityUpdateInput
+	): Promise<TechnologyEntity> {
+		const entity = await this.prismaClient.technologyEntity.update({
+			where: {
+				id,
+			},
+			data: updateTechnology,
+		});
+		await this.cache(entity);
```
mbicknese, line 51: What happens if Redis is unreachable at this point?

mbicknese: Although not what I was going for, that's interesting as well. Do we really want to block the whole application if we don't have caching set up? Could it do with a warning instead?

What I'm really after is, what happens if Redis dies while the stack is already running. Or what happens if there's a network hiccup and our Node instance fails to connect for this request?

mbicknese: Here's my concern about the code. The health endpoint will tell our monitoring whether our infrastructure is fully healthy. It does not alter how the production code behaves. Consider the following scenario:

 1. We start the stack, and everything is healthy
 2. After a while, Redis goes down. Our health endpoint reflects this
 3. In time, we try to get Redis back up
 4. Meanwhile, people will make requests which will call `updateTechnology` or `getTechnologyById`
 5. **if I'm not missing anything**, these requests will throw an error (500, most likely) because the Redis connection is dead.
 
So what I would like to know is whether we need to protect against this scenario. It would be a shame if our API goes down simply because we lost caching capabilities. 
Do you agree that we need a ticket for this?

starters/express-apollo-prisma/src/graphql/data-sources/technology-data-source.ts
```diff
@@ -0,0 +1,108 @@
...
+
+	async getTechnologies(): Promise<TechnologyEntity[]> {
+		return await this.prismaClient.technologyEntity.findMany();
+	}
+
+	async createTechnology(data: Prisma.TechnologyEntityCreateInput): Promise<TechnologyEntity> {
+		const entity = await this.prismaClient.technologyEntity.create({
+			data,
+		});
+		this.cache(entity);
```
mbicknese, line 37: This line seems to miss `await`

starters/express-apollo-prisma/.editorconfig
```diff
@@ -0,0 +1,16 @@
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+indent_style = tab
+
+[*.yml]
```
mbicknese, line 10:
```suggestion
[*.{yml,yaml}]
```

iansamz (approved):
LGTM

starter-kits.json
```diff
@@ -4,6 +4,7 @@
   "cra-rxjs-styled-components": "Create React App, RxJS and Styled Components",
   "deno-oak-denodb": "Deno, Oak, and DenoDB",
   "expo-zustand-styled-component": "Expo, Zustand, and Styled Components",
+  "express-apollo-prisma": "Express, Apollo and Prisma",
```
mbicknese, line 7:
```suggestion
  "express-apollo-prisma": "Express.js, Apollo Server, and Prisma",
```

mbicknese: Not because I need it to be those exact texts, but it's more inline with what we have in the other descriptions ;)

starters/express-apollo-prisma/package.json
```diff
@@ -0,0 +1,89 @@
+{
+	"name": "express-apollo-prisma",
+	"version": "0.0.1",
+	"description": "Express, Apollo and Prisma",
```
mbicknese, line 4: Regarding my earlier comment, I guess you need to change it here and run the generate kits command.

mbicknese (commented):
We're ready to merge it. Just a small remark on the description.

mbicknese (dismissed):
Awesome work @ihardz, @iansamz and @kodejuice !

mbicknese (approved):

PR #925, opened 2023-01-30
fix counter test

Megio:
Fix counter tests

### DIFF
--- a/starters/solidstart-tanstackquery-tailwind-modules/src/components/CounterExample/Counter.spec.tsx
+++ b/starters/solidstart-tanstackquery-tailwind-modules/src/components/CounterExample/Counter.spec.tsx
@@ -19,14 +19,16 @@ describe('CounterExample', () => {
     await render(() => <CounterExample />);
     const button = await screen.getByText('Decrement');
     fireEvent.click(button);
-    await expect(screen.getByText('Count: -1')).toBeDefined()
+    await expect(screen.getByText('Count: 0')).toBeDefined();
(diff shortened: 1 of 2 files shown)


starters/solidstart-tanstackquery-tailwind-modules/src/components/CounterExample/Counter.spec.tsx
```diff
@@ -19,14 +19,16 @@ describe('CounterExample', () => {
...
+    fireEvent.click(button);
+    await expect(screen.getByText('Count: -1')).toBeDefined();
   });
 
   it('should mount and reset', async () => {
     await render(() => <CounterExample />);
     const DecrButton = await screen.getByText('Decrement');
     fireEvent.click(DecrButton);
-    await expect(screen.getByText('Count: -1')).toBeDefined()
+    await expect(screen.getByText('Count: -2')).toBeDefined()
```
hdJerry, line 31: There is no need for the decrement btn again. just the rest test is okay since we have test for decrement already

hdJerry (changes requested):

starters/solidstart-tanstackquery-tailwind-modules/src/components/CounterExample/Counter.spec.tsx
hdJerry: And not sure why await is in front of expect. 
Please remove. It should still work

hdJerry (approved):
You Rock Sir!!

PR #976, opened 2023-02-08
[express-apollo-prisma] Create scalable caching solution

### LINKED ISSUE #917: [express-apollo-prisma] Create scalable caching solution
This is a follow-up issue to #907

The current solution for caching is hardcoded into the Technology resolver. This pattern can be copied and pasted for new models, but this doesn't scale.

One of the options we can consider is using Apollo's built-in caching. If it offers the required features for the kit, we can leverage the shipped solution.

If the Apollo caching is not the solution, we will need to abstract the caching solution. A quick search shows a package for prisma caching middleware.

ihardz:
## Type of change

- [ ] Documentation change
- [ ] Bug fix

## Summary of change

CacheAPIWrapper (Specification)

## Checklist

- [ ] This fix resolves #917

### DIFF
--- a/starters/express-apollo-prisma/src/graphql/schema/hello/hello.spec.ts
+++ b/starters/express-apollo-prisma/src/graphql/schema/hello/hello.spec.ts
@@ -1,7 +1,7 @@
 import gql from 'graphql-tag';
 import { Query } from '../generated/types';
 import assert from 'assert';
-import { serverExecuteOperation } from '../../utils/test';
+import { testServerExecuteOperation } from '../../utils/test';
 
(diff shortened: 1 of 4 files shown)


starters/express-apollo-prisma/package.json
```diff
@@ -12,7 +12,8 @@
 	"license": "MIT",
 	"main": "src/main.ts",
 	"scripts": {
-		"test": "jest",
+		"test": "npm run prisma:generate && npm run codegen && jest",
+		"dev:test": "jest",
```
iansamz, line 16: I think we can just `test` command is enough for both dev and production.

iansamz (approved):

PR #978, opened 2023-02-09
[serverless-apollo-contentful] Queueing feature unit tests

### LINKED ISSUE #963: [serverless-apollo-contentful] Queueing feature unit tests
# Description
- we need unit tests for each file. All files should have a corresponding test or set of tests.
- `feat/863-add-queueing-test` is the branch of work in progress

ihardz:
## Type of change

- [ ] Documentation change
- [ ] Bug fix

## Checklist

- [ ] This fix resolves #963

### DIFF
--- a/starters/graphql-serverless-contentful/src/handlers/sqs-generate-job.test.ts
+++ b/starters/graphql-serverless-contentful/src/handlers/sqs-generate-job.test.ts
@@ -6,8 +6,7 @@ import {
 } from 'aws-lambda';
 import { sendMessage } from '../utils/sqs/sendMessage';
 import { handler } from './sqs-generate-job';
-
-// jest.mock('@serverless-template/aws');
+import assert from 'assert';
 
 describe('generateDemoJob', () => {
 	let subject: void | APIGatewayProxyResult;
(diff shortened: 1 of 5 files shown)


starters/graphql-serverless-contentful/src/handlers/sqs-generate-job.test.ts
```diff
@@ -0,0 +1,83 @@
...
+				{} as Callback
+			);
+		});
+
+		afterAll(() => {
+			sendMessageMock.mockClear();
+		});
+
+		it('returns a 200 status code', () => {
+			if (subject) expect(subject.statusCode).toBe(200);
```
mbicknese, line 47: `subject` should be set here. If it's not set, `handler` is not behaving as expected.

Consider the scenario where something goes wrong and `subject` is not set. The if-statement will skip the assertion, and Jest will consider the case handled. It might give a warning for the empty test, but IIRC, it will not fail.

We can either drop the if-statement. Or, if this is a TypeScript thing, we can use `assert` to ensure `subject` is set and tell Jest when it's not.

```javascript
import assert from 'assert';

it('now uses assert! 🎉', () => {
  assert(subject != null); // type unsafe check to match all null-y values
  // TypeScript is now happy
})
```

starters/graphql-serverless-contentful/src/handlers/sqs-generate-job.test.ts
```diff
@@ -0,0 +1,83 @@
...
+		});
+
+		it('returns the returned messaged', () => {
+			if (subject)
```
mbicknese, line 51: Multiline if-statements without brackets scare me 😅 
Could you add brackets?

starters/graphql-serverless-contentful/README.md
```diff
@@ -147,7 +147,7 @@ How to run migrations:
    contentful space create --name "ThisDot example"
 ```
 
-- Set the newly-created space as the default space for all further CLI operations. This will present a list of all available spaces – choose the one we just created
+- Set the newly-created space as the default space for all further CLI operations. This will present a list of all available spaces � choose the one we just created
```
mbicknese, line 150: It looks like a weird character encoding issue happened here.

PR #1015, opened 2023-02-10
update custom issue template backend tickets

kenslachtajr:
## Type of change

- [ ] Documentation change
- [ ] Bug fix

## Checklist

- [ ] This fix resolves #

### DIFF
--- a/.github/custom-issue-template/backend-kit/1. Add linting and prettier configurations.md
+++ b/.github/custom-issue-template/backend-kit/1. Add linting and prettier configurations.md
@@ -2,6 +2,11 @@
 
 Having code style consistency is important for any project. A standardized set of options should be set that can be overwritten by the implementer at a later date.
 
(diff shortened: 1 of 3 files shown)


.github/custom-issue-template/backend-kit/1. Add linting and prettier configurations.md
```diff
@@ -2,6 +2,11 @@
 
 Having code style consistency is important for any project. A standardized set of options should be set that can be overwritten by the implementer at a later date.
 
+- ESLint: Code formatting should be enforced via ESLint rules. Rules should be set to the strictest setting and enforced via CI. The standard rules for a stack should be selected and if there are specialized plugins, those should be utilized and appropriately tuned.
```
Zelig880, line 5: same comment as the FE ticket changes. i think this is too strict.

Zelig880 (changes requested):
the comment here is linked to what we agree on the FE ticket

.github/custom-issue-template/backend-kit/1. Add linting and prettier configurations.md
kenslachtajr: These are taken directly from the Specification. If we want to make this less strict, we should probably visit this in the Specification.

image

PR #1023, opened 2023-02-10
[serverless-apollo-contentful] Fix: Refactor Unit tests: healthcheck.test.ts

### LINKED ISSUE #1017: [serverless-apollo-contentful] Fix: Refactor Unit tests: healthcheck.test.ts
Description: mock real external instances

ihardz:
## Type of change

- [ ] Documentation change
- [ ] Bug fix

## Summary of change

Constains implementations refactoing, fixes and missing funcitonality

## Checklist

- [ ] This fix resolves #1017


starters/serverless-apollo-contentful/src/utils/contentful/check-health.ts
```diff
@@ -0,0 +1,23 @@
+import { QueryOptions } from 'contentful-management';
+import { CONTENTFUL_CONTENT_TYPE_ID_TECHNOLOGY } from '../../contentful/constants';
+import { getEnvironment } from './get-environment';
+
+export const CONTENTFUL_HEALTH_CHECK_QUERY_OPTIONS: QueryOptions = {
+	content_type: CONTENTFUL_CONTENT_TYPE_ID_TECHNOLOGY,
```
mbicknese, line 6: I'm not a big fan of tying our content model to the health check like this. My concern is about the developer experience. Users of the kit will probably delete the `technology` model. This will "suddenly" cause the health endpoint to fail, potentially making them look for connection issues (e.g., ensuring their API key is correctly set). It also raises the question of what should happen when there's more than one model? Do we test them all? Do you pick one, and if that's fine, the whole connection is OK?

A simple solution might be verifying if we can get the environment without error. Supposedly that authenticates against contentful, and if that succeeds, the connection should be in working order. I'm not sure if this is how the client works under the hood, though.

If that is not working as expected. Possibly Contentful offers other queries that are not bound to the models in the application. Can we query which content types exist in our environment?

starters/serverless-apollo-contentful/src/utils/contentful/constants.ts
```diff
@@ -0,0 +1 @@
+export const CONTENTFUL_CONTENT_TYPE_ID_TECHNOLOGY = 'technology';
```
mbicknese, line 1: Shall we make this part of the model?

e.g.,
```typescript
class TechnologyModel {
  public static readonly NAME = 'technology';
}

mbicknese (changes requested):

mbicknese (changes requested):
I've removed the contentful health check, so it's not blocking this feature.

The healthcheck handler test seems incomplete. Line 27 is invalid and the implementation of the tests seems te be missing. I guess that's why the PR is set to WIP. I just wanted to make sure it's known.

starters/serverless-apollo-contentful/src/utils/contentful/constants.ts
mbicknese: Alright, lets keep it as is.

Issue #1002, opened 2023-02-09
Starter kit package update (Remix-gql-tailwind) - Minor updates/patches only


WillHutt:
# Background

This ticket aims to update the Remix-gql-tailwind starter kit to address out-of-date package patch cycles.

**Only do yellows and greens.**

**Don't do major upgrades with breaking changes or changes that would change the identity of the kit, prioritize bringing minor patches up to current.**

![Image](https://user-images.githubusercontent.com/20880360/217929959-bb53a359-fea4-4c0b-a97d-d754f7563ffa.png)

# Acceptance
- [ ] Update yellow and green packages
- [ ] Verify everything still works as expected

PR #1037, opened 2023-02-13
chore(remix-gql-tailwind): Minor version dep upgrades

danecando:
Upgrades dependencies to latest minor versions

Closes https://github.com/thisdot/starter.dev/issues/1002


jdwilkin4 (commented):
I ran it locally with the updates and it works fine on my end. 
I did have a question about these empty values here

```
  --tw-pinch-zoom:  ;
```
Do we need all of those variables?

danecando:
> I ran it locally with the updates and it works fine on my end. I did have a question about these empty values here
> 
> ```
>   --tw-pinch-zoom:  ;
> ```
> 
> Do we need all of those variables?

I would guess no since it has no value, but it looks like it's stuff that is generated by the tailwind compiler

jdwilkin4 (approved):
Happy to approve 👍

PR #1153, opened 2023-03-02
Feature/790 generator metrics tracking

### LINKED ISSUE #790: [Feature]: Allow ThisDot to gather statistics on generator usage
### Is your feature request related to a problem? Please describe.

_As the product owner, I'd like to be able to see statistics on how used our generator tool is, and which kits are the most used._

**Additional problems to consider:**
- Can we gather these kinds of statistics? We don't want to have any data from the user, but is it legal to just gather this information?
- If we can send requests out, where should we store the data?

**Storing the data**
We can build our own infrastructure (a serverless function with dynamodb connection) or we could also use a third-party metrics system.

If we build our own infrastructure, it would be an additional feature to make a UI where the statistics can be viewed. If we go with a third-party metrics system (e.g.: Mixpanel), that also has costs but comes with everything we might need.

### Describe the solution you'd like

The generator script has room for sending a request out to a hardcoded endpoint with a payload that can help us track information on which kits get used the most.

It might also pose a challenge, since native fetch is only supported on later nodejs releases, therefore, we might need to bundle a package which we could use for making the request (e.g.: axios).

If we use a 3rd party metrics solution, their client might has this solved. Either way every kind of error should be suppressed, so kit generation is not affected by our metrics information gathering.

### Describe alternatives you've considered

GitHub Insights might give some insights on usage, but I don't know if it could be fine-tuned to detect which folders got cloned.

### Additional context

_No response_

TapaiBalazs:
## Type of change

- [X] Feature
- [ ] Documentation change
- [ ] Bug fix

## Summary of change

- Added a metrics tracking call to the generator
- Added a metrics-handler serverless project as a package to the repository

## Checklist

- [X] These changes follow the [contributing guidelines](https://github.com/thisdot/starter.dev/blob/main/CONTRIBUTING.md)
- [X] This feature has been approved by the maintainers
- [X] All dependencies are version locked
- [X] This fix resolves #790 


packages/create-starter/src/index.ts
```diff
@@ -56,6 +57,8 @@ export async function main() {
     process.exit(1);
   }
 
+  trackSelectedKit(options.kit)
```
dustinsgoodman, line 60: You should `await` this. If the CLI exits before the async call, it will abort and we lose the data.

packages/metrics-handler/index.js
```diff
@@ -0,0 +1,57 @@
...
+  if (!starterKitsRequest.ok) {
+    console.log('Could not fetch starter kits for validation, tracking aborted');
+    return responseHelper('Could not fetch starter kits for validation, tracking aborted')
+  }
+
+  const starterKitsJSON = await starterKitsRequest.json();
+  const starterKits = new Set()
+  if (typeof starterKitsJSON === 'object' && starterKitsJSON !== null) {
+    Object.keys(starterKitsJSON).forEach((kit) => starterKits.add(kit))
+  }
```
ktrz-sre-is, line 30: There is no need to iterate manually. Set constructor accepts iterable as a parameter
```suggestion
  const starterKits = new Set(
    typeof starterKitsJSON === 'object' && starterKitsJSON !== null
      ? Object.keys(starterKitsJSON)
      : []
  )
```

packages/metrics-handler/index.js
```diff
@@ -0,0 +1,57 @@
+function responseHelper(body) {
```
ktrz-sre-is, line 1: this name is not communicating the intention very well. Maybe let's call it `createOkResponse` or something similar?

packages/create-starter/src/index.ts
```diff
@@ -56,6 +57,8 @@ export async function main() {
     process.exit(1);
   }
 
+  await trackSelectedKit(options.kit)
```
ktrz-sre-is, line 60: I don't think it's gonna be a good user experience to force the CLI to wait for the tracking before it's gonna go with the cloning of the starter kit. Maybe we could extract the remainder of the function after this call into a separate function and call them in parallel?

```suggestion
  const [createStarterResult] = await Promise.allSettled([
    createStarter(/* necessary params */), 
    trackSelectedKit(options.kit)
  ])
  if (createStarterResult.status === 'rejected') {
    const err = createStarterResult.reason
    console.log(red(err instanceof Error ? err.message : 'Creating starter kit failed')))
    process.exit(1)
  }
```

packages/metrics-handler/index.js
ktrz, line 30: There is no need to iterate manually. Set constructor accepts iterable as a parameter
```suggestion
  const starterKits = new Set(
    typeof starterKitsJSON === 'object' && starterKitsJSON !== null
      ? Object.keys(starterKitsJSON)
      : []
  )```

File: packages/metrics-handler/index.js [hunk 5]
```diff
@@ -0,0 +1,57 @@
...
+  return {
+    statusCode: 200,
+    body: JSON.stringify(body),
+  };
+}
```
ktrz, line 6: this name is not communicating the intention very well. Maybe let's call it `createOkResponse` or something similar?

packages/create-starter/src/index.ts
ktrz, line 60: I don't think it's gonna be a good user experience to force the CLI to wait for the tracking before it's gonna go with the cloning of the starter kit. Maybe we could extract the remainder of the function after this call into a separate function and call them in parallel?

```suggestion
  const [createStarterResult] = await Promise.allSettled([
    createStarter(/* necessary params */), 
    trackSelectedKit(options.kit)
  ])
  if (createStarterResult.status === 'rejected') {
    const err = createStarterResult.reason
    console.log(red(err instanceof Error ? err.message : 'Creating starter kit failed')))
    process.exit(1)
  }```

packages/metrics-handler/index.js
dustinsgoodman: 😂 I wrote this and it's something I use in my other projects. It usually accepts a status and body but for this use case, we only want to send 200s. I think the rename is fine but just sharing the history.

packages/create-starter/src/index.ts
dustinsgoodman: Oh, that's better. Originally, the request wasn't awaited but if the process exited before the request finished, we would lose metrics. This version definitely clears up that problem. @TapaiBalazs had the same concern as you but I said this request would be negligible enough to afford the `await`. This compromise is much better. 😂

packages/create-starter/src/index.ts
TapaiBalazs: done

packages/metrics-handler/index.js
TapaiBalazs: renamed to `createOkResponse`

packages/create-starter/src/index.ts
TapaiBalazs: done

dustinsgoodman (approved):

PR #1160, opened 2023-03-08
[serverless-framework-apollo-contentful] Fixes and reworks for a polished kit

### LINKED ISSUE #1154: [serverless-apollo-contentful] Address review feedback
- [x] It's Serverless Framework, not AWS Serverless. We should brand things correctly.

- [x] The kit on the CLI is called "AWS Serverless, GraphQL Apollo Server, and Contentful CMS" but on the website, I don't see Contentful and see GraphQL listed twice.

- [x] The .nvmrc file has a specific node version specified. We should list it as a LTS version instead. I believe Node 16 is lts/gallium

- [x] README says to run infrastructure:start after installing but you need to copy the .env.example to .env first

- [x] First run of dev and I have warnings that should not be there. One is complaining about the iamRoleStatements in the serverless yml for the job function. You can't set function level iamRoleStatements like this without a plugin. You need to set them at the provider level which I would recommend doing. The rest appear to be typescript warnings.

Image

Image

- [x] Errors are coming back as JSON strings and not objects. Probably should be parsing these better.

Image

- [x] There is no "Get All" query for technologies

- [x] updateTechnology doesn't quite follow best practices here. You don't want to make all the fields top level - it can make more complicated object inputs much deeper. It's considered better practice to create a custom input to wrap this. I would expect this to be updateTechnology(id: ID!, fields: TechnologyUpdateFields) where the input is defined as:

Image
Note how all fields are optional at this level

Image

- [x] Good job on testing!  If we can get branch % to over 80% then that'd be great

Image

- [x] linting errors

Image

- [x] format causes writes

Image

- [x] Using class models is a code smell in serverless infrastructures. I have a very long write-up about this that you should read but TechnologyModel.ts is fighting best practices here and probably should be broken up differently https://medium.com/@dustinsgoodman/resolving-serverless-webpack-issues-efae729e0619

- [x] This function in resolvers is very anti-practice GraphQL. Instead you should define the Technology key in resolvers and set each of these different mappers to a resolver function, see 2nd screenshot. You could also handle this at the model fetch layer which might be the best. 

Image

Image

- [x] printWidth set to 80 in the prettier config is a bit aggressive. 100 or 120 is pretty standard these days.

- [x] serverless.yml best practices:

Image

mbicknese:
## Type of change

- [ ] Documentation change
- [ ] Bug fix

- [x] All of the above

## Summary of change

```
----------------------------|---------|----------|---------|---------|-------------------
File                        | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
----------------------------|---------|----------|---------|---------|-------------------
All files                   |   94.44 |    80.55 |   93.33 |   95.34 |                   
 handlers                   |     100 |      100 |     100 |     100 |                   
  healthcheck.ts            |     100 |      100 |     100 |     100 |                   
  sqs-generate-job.ts       |     100 |      100 |     100 |     100 |                   
  sqs-handler.ts            |     100 |      100 |     100 |     100 |                   
 models/Technology          |   84.78 |    73.91 |   85.71 |   88.09 |                   
  TechnologyModel.ts        |   81.81 |    73.91 |      80 |   89.47 | 23,30             
  create.ts                 |     100 |      100 |     100 |     100 |                   
  getAll.ts                 |   72.72 |      100 |     100 |      70 | 13-15             
  getById.ts                |     100 |      100 |     100 |     100 |                   
 schema/technology          |     100 |      100 |     100 |     100 |                   
  technology.resolvers.ts   |     100 |      100 |     100 |     100 |                   
 utils/contentful           |     100 |      100 |     100 |     100 |                   
  contentful-healthcheck.ts |     100 |      100 |     100 |     100 |                   
  contentful.ts             |     100 |      100 |     100 |     100 |                   
 utils/redis                |     100 |      100 |     100 |     100 |                   
  redis-healthcheck.ts      |     100 |      100 |     100 |     100 |                   
 utils/sqs                  |   97.61 |       80 |     100 |   97.29 |                   
  client.ts                 |    92.3 |       75 |     100 |   91.66 | 9                 
  getQueueUrl.ts            |     100 |      100 |     100 |     100 |                   
  is-offline.ts             |     100 |      100 |     100 |     100 |                   
  sendMessage.ts            |     100 |      100 |     100 |     100 |                   
----------------------------|---------|----------|---------|---------|-------------------

Test Suites: 16 passed, 16 total
Tests:       65 passed, 65 total
```

## Checklist

- [x] This fix resolves #1154


iansamz (approved):
LGTM

PR #1170, opened 2023-03-10
feat: ESLint and Prettier configs

danecando:
## Changes

- Adds some placeholder options as a starting point for ESLint configs
- Adds Prettier and basic prettier config (same as other Next 12 kit)
- Adds missing LICENSE file
- Turns on reactStrictMode in NextJS config


jdwilkin4 (approved):
looks good 👍

PR #1179, opened 2023-03-13
Update mh-deploy.yml to fix deploy

BigAB:
I am not sure this is the right change, because I actually don't fully understand what is being configured with the `working-directory` element

...but it makes the squiggly lines go away! 😄

## Type of change

- [x] Bug fix

## Summary of change

I just changed the indentation of the yml


dustinsgoodman (approved):
Gotta love yaml alignment issues.

lindakatcodes (approved):

PR #1185, opened 2023-03-15
fix(next13-zustand-bulma): remove extraneous count.ts file (merge error)

BigAB:
## Type of change

So this just removes a duplicate `count.ts` file, that I think came from all my rebasing tomfoolery

- [ ] Documentation change
- [x] Bug fix

## Checklist

- [X] I have verified the fix works and introduces no further errors


lindakatcodes (approved):
We've all been there. lol 😊

PR #1186, opened 2023-03-15
fix(next13-zustand-bulma): Bulma tag class on storybook Source component

BigAB:
Storybook's `Source` component has a `class="tag"` and therefore Bulma's `tag` class is being picked up.

This PR adds a hacky little fix to ignore `.tag` if it's the ancestor of a `pre` element

## Type of change

- [x] Bug fix

### Before
Screenshot 2023-03-15 at 4 08 35 PM

### After
Screenshot 2023-03-15 at 4 03 37 PM

## Checklist

### DIFF
--- a/starters/next13-zustand-bulma/.storybook/main.ts
+++ b/starters/next13-zustand-bulma/.storybook/main.ts
@@ -14,5 +14,27 @@ const config: StorybookConfig = {
   docs: {
     autodocs: 'tag',
   },
+  /* This is a little hack to fix the issue with Bulma's tag
+   component from the `tag` class being picked up by storybooks
+   `Source` component and messing with the styles.
+  */
+  previewHead: (head) => `
+  ${head}
+  <style>
+    pre.prismjs .tag {
+      align-items: initial;
+      background-color: initial;
+      border-radius: initial;
+      display: initial;
+      font-size: initial;
+      height: initial;
+      justify-content: initial;
+      line-height: initial;
+      padding-left: initial;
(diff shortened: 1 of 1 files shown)


starters/next13-zustand-bulma/src/app/globals.scss
```diff
@@ -16,3 +16,21 @@
...
+  border-radius: initial;
+  display: initial;
+  font-size: initial;
+  height: initial;
+  justify-content: initial;
+  line-height: initial;
+  padding-left: initial;
+  padding-right: initial;
+  white-space: initial;
+}
```
william-mimura-thisdot, line 36: My thought was to add this only to the '.storybook/preview.ts` so this would only affect the storybook, just in case the user wants to create a `<pre>` + `<tag>` tags and still keep the Bulma styles in their project.

I'm happy to approve this if there's a reason to do this here though. 👍

BigAB: Okay I moved it to the `.storybook/main.ts` (because I didn't know how to add it to `preview.ts`) and I agree it's much better placed there.

The problem is, that these components (namely Source) aren't really following the storybook rules of what is "manager" and "preview" and so the styles are mixed, because they are both in the "preview" and I am unsure how to limit it only to the storybook Source component.

(if you have ideas, I am all on board)

By adding the extra `.prismjs` class hopefully it's narrow enough to not cause problems.

william-mimura-thisdot (approved):
LGTM!

PR #1200, opened 2023-03-22
Fix/1197 address demo comments

### LINKED ISSUE #1197: [serverless-apollo-contentful] Final cleanup based on demo feedback
# Description 
After our demo of the kit, there are a few minor items of cleanup that need to take place. They are as follows:

- Change name to "Serverless Framework"
-

mbicknese:
## Type of change

- [ ] Documentation change
- [x] Bug fix

## Summary of change

- Change name to "Serverless Framework"
- Remove double quotes from .env file AWS credentials
- Update README to include the fact that both migrations exist
- Does the Contentful system not return pagination by their API? If pagination isn't available, note in README. If so, refactor to use it.
- Check and confirm that the mutations (update) are working correctly. There was a potential bug during the demo.
- In the graphql.ts file, SLS deploy use --stage flag for Serverless Framework
- Remove left-over console log from debugging
- Update tests so no errors are returned

## Checklist

- [x] This fix resolves #1197 
- [x] Test coverage is over 80%

### DIFF
--- a/starters/serverless-framework-apollo-contentful/.env.example
+++ b/starters/serverless-framework-apollo-contentful/.env.example
@@ -14,6 +14,6 @@ REDIS_CACHE_TTL_SECONDS=900
 SQS_PORT=9324
 
 # Allows you to run Serverless SQS offline
-AWS_ACCESS_KEY_ID="offline"
-AWS_SECRET_ACCESS_KEY="offline"
-AWS_SESSION_TOKEN="offline"
+AWS_ACCESS_KEY_ID=offline
+AWS_SECRET_ACCESS_KEY=offline
+AWS_SESSION_TOKEN=offline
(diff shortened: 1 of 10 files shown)


mbicknese:
I've not added "server" to the apollo part in the kit name. As I understood it this was not a requirement but would had been nice if we had done it right initially.

starters/serverless-framework-apollo-contentful/.env.example
```diff
@@ -14,6 +14,6 @@ REDIS_CACHE_TTL_SECONDS=900
 SQS_PORT=9324
 
 # Allows you to run Serverless SQS offline
-AWS_ACCESS_KEY_ID="offline"
-AWS_SECRET_ACCESS_KEY="offline"
-AWS_SESSION_TOKEN="offline"
+AWS_ACCESS_KEY_ID=offline
+AWS_SECRET_ACCESS_KEY=offline
+AWS_SESSION_TOKEN=offline
```
iansamz, line 19: The SLS_STAGE is missing here

iansamz (changes requested):

iansamz (approved):

PR #1208, opened 2023-03-31
Vue3 xstate kit

### LINKED ISSUE #518: [Vue3 + XState] Add full kit to the repo
Acceptance criteria: 

- [ ] Once all tasks are completed, open PR against main branch to merge in kit

lindakatcodes:
## Type of change

- [ ] Documentation change
- [ ] Bug fix
- [x] New kit

## Summary of change

Closes #518 

Creates a new starter kit using Vue 3, XState, and CSS. Includes Cypress component testing and Storybook along with the usual linting and formatting things.

## Checklist


starters/vue3-xstate-css/.prettierrc.json
```diff
@@ -0,0 +1,7 @@
+{
+  "singleQuote": true,
+  "semi": true,
+  "tabWidth": 2,
+  "trailingComma": "es5",
+  "useTabs": false
```
jdwilkin4, line 6: it looks like we might need to change this to "useTabs": true because in the spec it says this


It has been proven that tabs are better than spaces for accessibility purposes. As such, all tooling should be configured to use tabs when possible. YAML is a known exception.

starters/vue3-xstate-css/README.md
```diff
@@ -0,0 +1,196 @@
...
+
+- [Node.js 16.8](https://nodejs.org/) or later installed
+
+### Development
+
+Once you've generated your new starter kit following the instructions above:
+
+- `cd` into your project directory and run `npm install` to install the dependencies.
+- Run `npm run dev` to start the development server.
+- Click the [http://localhost:5173] link from within your terminal. This will open that link in a tab to view the application, and also launch the XState inspector for you in a second tab.
```
jdwilkin4, line 48: small markdown issue here

Screenshot 2023-03-31 at 4 46 04 PM

```suggestion
- Click the [http://localhost:5173](http://localhost:5173) link from within your terminal. This will open that link in a tab to view the application, and also launch the XState inspector for you in a second tab.
```

jdwilkin4 (commented):
starting reviewing this and found a couple of small things but looks good so far 👍

starters/vue3-xstate-css/src/views/CounterView.vue
```diff
@@ -0,0 +1,51 @@
...
+import ButtonComponent from '../components/ButtonComponent.vue';
+import HeaderComponent from '../components/HeaderComponent.vue';
+import { counterMachine } from '../machines/counterMachine';
+
+const { state, send } = useMachine(counterMachine, { devTools: true });
+</script>
+
+<template>
+  <main>
+    <HeaderComponent>
```
jdwilkin4, line 12: For mobile devices, the layout would work better in a column here

Screenshot 2023-04-05 at 12 00 08 PM

starters/vue3-xstate-css/src/views/GreetView.vue
```diff
@@ -0,0 +1,78 @@
...
+      <div class="fetch__section-result" data-cy="message-result">
+        <div
+          v-if="state.value === 'loading'"
+          class="fetch__section-loader"
+        ></div>
+        <p v-if="state.value === 'failure'" class="fetch__section-message-fail">
+          {{ state.context.error }}
+        </p>
+        <p v-else class="fetch__section-message-success">
+          {{ state.context.message }}
```
jdwilkin4, line 29: For mobile, I think reducing the font size so the text is all in one line would work better here
Screenshot 2023-04-05 at 12 03 02 PM

starters/vue3-xstate-css/src/machines/greetMachine.ts
```diff
@@ -0,0 +1,55 @@
...
+      },
+      initial: 'loading',
+      context: {
+        query,
+        message: '',
+        error: '',
+      },
+      states: {
+        loading: {
+          invoke: {
```
jdwilkin4, line 25: When I run `yarn run type-check ` I receive this error message 

Screenshot 2023-04-05 at 12 13 36 PM

starters/vue3-xstate-css/src/stories/ButtonComponent.stories.js
```diff
@@ -0,0 +1,19 @@
+import ButtonComponent from '../components/ButtonComponent.vue';
```
jdwilkin4, line 1: Is this the default story that automatically comes with storybook?
If so, I think we should remove it because the spec mentions removing all default stories.

starters/vue3-xstate-css/src/stories/HeaderComponent.stories.js
```diff
@@ -0,0 +1,19 @@
+import HeaderComponent from '../components/HeaderComponent.vue';
```
jdwilkin4, line 1: Is this the default story that automatically comes with storybook?
If so, I think we should remove it because the spec mentions removing all default stories.

jdwilkin4 (dismissed):
I was able to finish going through everything and testing out the site as well as all of the commands.
Looks good, left a few comments 👍

starters/vue3-xstate-css/README.md
lindakatcodes: Dang, the direct link with <> was causing crashes but I thought this would work without specific text. Good catch!

starters/vue3-xstate-css/src/views/GreetView.vue
lindakatcodes: What about putting "Message:" on one line and the response on another? I hesitate to make text too small on mobile, though even in one line it wouldn't have to be that much smaller I don't think. I can play with this!

starters/vue3-xstate-css/src/machines/greetMachine.ts
lindakatcodes: I may need a comment here for this, that's a good point. This is exactly how the XState docs have it and from all I could find it should be supported, but I couldn't ever get this issue to clear. I'm hoping I can eventually share this kit and ask in their Discord about it, maybe I have something weird in my types. But I can add a comment for now to explain that.

starters/vue3-xstate-css/src/stories/ButtonComponent.stories.js
lindakatcodes: Nope this is a specific one I made for this kit. :)

starters/vue3-xstate-css/src/stories/HeaderComponent.stories.js
lindakatcodes: Nope this is a specific one I made for this kit. :)

starters/vue3-xstate-css/src/views/GreetView.vue
jdwilkin4: I like the idea of having message and the response on separate lines

lindakatcodes:
I had to add a `@ts-expect-error` call in the `greetMachine` function because VS Code isn't playing nice. I've made any adjustments I've been able to search online and directly comparing my code line by line with other examples the usage of invoke is correct, but the type refuses to use the state invoke type and not the overall invoke type. 

This issue is the closest I've found to what I've been seeing: https://github.com/statelyai/xstate/issues/986

I also set up this code sandbox example from the one David links in the above ticket - I've left his code there commented out for testing, and copied over my machine directly as is (the only exception is I also directly copied over the service function instead of importing it, but testing locally that difference didn't matter). In the sandbox I do NOT get the error, even with the matching XState and TS versions. I also added in the tsconfig files, the `@types/node` and vite packages as those were the only things I could think might be affecting the types, but still no error in the sandbox but consistent error in my editor. 🤷🏼‍♀️ 
https://codesandbox.io/s/eager-hill-6wgmtx?file=/src/index.ts

starters/vue3-xstate-css/src/router/index.ts
```diff
@@ -0,0 +1,27 @@
+import { createRouter, createWebHistory } from 'vue-router';
+
+const router = createRouter({
+	history: createWebHistory(import.meta.env.BASE_URL),
+	routes: [
+		{
+			path: '/',
+			name: 'home',
+			component: () => import('../views/HomeView.vue'),
```
Zelig880, line 9: You can leave it as it is, but it is not suggested to "async load" the root of a website as you know that it will mostly always be loaded.

I know that for this size project it is ok, but it is a nice to have

starters/vue3-xstate-css/src/App.vue
```diff
@@ -0,0 +1,3 @@
+<template>
+	<RouterView />
```
Zelig880, line 2: I would actually wrap this in MAIN, so that all the view loaded will be within main sop:

```
<template>
  <main>
    <RouterView />
  </main>
</tempalte>
```

starters/vue3-xstate-css/src/views/HomeView.vue
```diff
@@ -0,0 +1,35 @@
+<script setup lang="ts"></script>
+
+<template>
+	<main>
+		<div class="home__title">
+			<h1>Vue 3 + XState + CSS</h1>
+		</div>
+		<div class="home__links">
```
Zelig880, line 8: small accessibilty suggestion, make this a <nav> instead than a div

Zelig880 (changes requested):
Great code!!

i did not run the app locally and just provided comments from looking at the code.

starters/vue3-xstate-css/src/views/HomeView.vue
lindakatcodes: Looks like something went weird with the text - what should this be? A section maybe?

Zelig880: A nav

starters/vue3-xstate-css/src/App.vue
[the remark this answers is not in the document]
lindakatcodes: Nice catch - the prettier file is actually set to use tabs instead of spaces, but the editor config never got updated along with that. So I'll fix this in the editor config!

starters/vue3-xstate-css/src/machines/counterMachine.ts
```diff
@@ -0,0 +1,32 @@
...
+			},
+		},
+	},
+	{
+		actions: {
+			increment: assign({ count: (context) => context.count + 1 }),
+			decrement: assign({ count: (context) => context.count - 1 }),
+			reset: assign({ count: (context) => (context.count = 0) }),
+		},
+	}
```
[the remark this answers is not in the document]
lindakatcodes: When I try to re-situate it like this, I'm getting type errors for `on`, and nothing I'm finding in the docs is helping to alleviate that.

But also, from this section of the docs on actions, they recommend declaring the actions in the config object that's passed as the second object, which is how I have it here already. So despite slightly conflicting info, I think the way it currently is might actually be a better practice overall.
Screen Shot 2023-05-05 at 6 55 19 PM
https://xstate.js.org/docs/guides/actions.html#api

The condition isn't set because in the starter specifications, it says it should be able to work in the negative direction as well: https://docs.google.com/document/d/1iEXK6BFgu5HZ4dfU7pfv80zLOAd86T8AzstCMEYTUXA/edit#heading=h.o7at9s7kpri4

starters/vue3-xstate-css/src/router/index.ts
```diff
@@ -0,0 +1,27 @@
...
+			name: 'counter',
+			component: () => import('../views/CounterView.vue'),
+		},
+		{
+			path: '/greet',
+			name: 'greet',
+			component: () => import('../views/GreetView.vue'),
+			props: {
+				query: 'from This Dot Labs!',
+			},
```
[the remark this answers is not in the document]
lindakatcodes: Turns out Vue 3 includes a way to provide and inject values, so I swapped this out to use that functionality! The value is now provided by the app so it's available to any children, and then the GreetView component can inject the provided value and use an empty string as a fallback value (so it becomes optional).

starters/vue3-xstate-css/src/tests/GreetView.cy.ts
```diff
@@ -0,0 +1,36 @@
...
+	it('should display a message with my greeting when I pass in a greeting', () => {
+		cy.mount(GreetView, { props: { query: 'from Cypress!' } });
+		cy.get('[data-cy=message-result]').should(
+			'contain.text',
+			'Hello, from Cypress!'
+		);
+	});
+
+	it('should display a default message when no greeting is passed', () => {
+		cy.mount(GreetView, { props: { query: '' } });
```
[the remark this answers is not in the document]
lindakatcodes: Managed to achieve this using Vue 3's provide and inject functionality. So we provide a value from the root of the app, but set a default empty string value when it's injected so if we didn't provide a value it will use an empty string instead.

starters/vue3-xstate-css/src/components/ButtonComponent.vue
```diff
@@ -0,0 +1,28 @@
+<script setup lang="ts"></script>
+
+<template>
+	<button>
+		<slot />
+	</button>
+</template>
+
+<style scoped>
+button {
```
AllanJeremy, line 10: I think we should add some type of hover state for accessibility here

AllanJeremy (changes requested):
LGTM. Left some minor comments

starters/vue3-xstate-css/src/router/index.ts
```diff
@@ -0,0 +1,25 @@
+import { createRouter, createWebHistory } from 'vue-router';
+import HomeView from '../views/HomeView.vue';
+
+const router = createRouter({
+	history: createWebHistory(import.meta.env.BASE_URL),
+	routes: [
+		{
+			path: '/',
+			name: 'home',
+			component: HomeView,
```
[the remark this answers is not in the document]
lindakatcodes: Lol I had this dynamic before, but Simone suggested it should be a direct import since the home page is basically always expected to exist. 
https://github.com/thisdot/starter.dev/pull/1208#discussion_r1185066650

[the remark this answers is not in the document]
lindakatcodes: Alright, after a bit of discussion and looking at the official Vue Router docs, I'm going to revert this and use the dynamic import. 

For reference if anyone is curious, in the [Vue Router docs](https://router.vuejs.org/guide/advanced/lazy-loading.html#lazy-loading-routes) they say "In general, it's a good idea to always use dynamic imports for all your routes." 

And per some internal discussion, if the root page is visited the component will need to be imported whether it's static or dynamic. But if someone uses this kit to make their own project, and they end up having other pages that are visited directly (ie a link to a blog post or something), it's feasible that some users won't actually need the home component loaded.

packages/website/src/icons/XstateIcon.tsx
```diff
@@ -0,0 +1,53 @@
...
+      viewBox="0 0 256 96"
+      version="1.1"
+      xmlns="http://www.w3.org/2000/svg"
+      className={className}
+    >
+      <g>
+        <g id="Layer_6">
+          <path
+            d="M55.8,10.4L55.8,10.4h19.5L55.6,44.6l0,0C49.6,33.9,49.7,20.9,55.8,10.4z"
+            fill="var(--fill, black)"
```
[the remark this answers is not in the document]
lindakatcodes: Oh good catch - I auto generated this so didn't catch the var.

PR #1241, opened 2023-05-02
[Lit-Mobx-Vaadin] Fetch Example

### LINKED ISSUE #1223: [Lit-Mobx-Vaadin] Create the Fetch Page and Component (with tests and storybook)

Amdrel:
## Type of change

- [ ] Documentation change
- [ ] Bug fix
- [x] Feature

## Summary of change

Adds the fetch example to the Lit starter.

## Checklist

- [x] This fix resolves #1223
- [x] I have verified the feature works and introduces no further errors

### DIFF
--- a/starters/lit-mobx-vaadin/src/components/fetch/fetch.ts
+++ b/starters/lit-mobx-vaadin/src/components/fetch/fetch.ts
@@ -1,13 +1,9 @@
-import { MobxLitElement } from '@adobe/lit-mobx';
-import { PropertyValueMap, css, html } from 'lit';
-import { customElement } from 'lit/decorators.js';
-import { starterState } from '../../state.js';
-
-const FETCH_API_URL = 'https://api.starter.dev/.netlify/functions/server/hello';
+import { LitElement, css, html } from 'lit';
+import { customElement, property } from 'lit/decorators.js';
 
 @customElement('td-fetch')
-export class StarterFetch extends MobxLitElement {
-  protected state = starterState;
+export class StarterFetch extends LitElement {
+  @property({ type: String }) message = '';
 
   static styles = [
     css`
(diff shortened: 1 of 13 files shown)


starters/lit-mobx-vaadin/src/pages/page.mixin.ts
```diff
@@ -2,19 +2,42 @@ import { LitElement, css } from 'lit';
...
+          text-align: center;
+        }
+
         h1 {
-            border-bottom: 5px solid var(--td-primary-color, var(--td-text-color, #1d4ed8));
+          padding: 15px;
+          font-size: 2rem;
+          font-weight: 700;
+          text-align: center;
+          border-bottom: 5px solid rgb(29 78 216);
```
flakolefluk, line 25: shouldn't all this colors become css variables?, i haven't checked all the colors, but if some are too similar, could it be that the reference is a bit inconsistent with coloring?

starters/lit-mobx-vaadin/src/components/fetch/fetch.ts
```diff
@@ -0,0 +1,87 @@
...
+    try {
+      const response = await fetch(
+        `${FETCH_API_URL}?${new URLSearchParams({ greeting })}`,
+      );
+      return response.text();
+    } catch (error) {
+      console.error(error);
+      return 'Sorry, something went wrong. Please try again.';
+    }
+  }
```
flakolefluk, line 86: I don't think we really need to async/await here, In practice I'd think we would have a service with requests, or an external file, for reusability. May be too much for this example, but this function could live outside the class anyway. I think this could also  help with the escaped linter rules ("class-methods-use-this": "off",)
```suggestion
  private fetchMessage(greeting: string): Promise<string> {
    return fetch(
      `${FETCH_API_URL}?${new URLSearchParams({ greeting })}`,
    ).then(res => res.text()).catch(error => {
      console.error(error);
      return 'Sorry, something went wrong. Please try again.';
    })
  }
```

starters/lit-mobx-vaadin/src/pages/counter-example/starter-counter.ts
```diff
@@ -12,6 +12,11 @@ export class StarterCounter extends PageMixin(LitElement) {
       <vaadin-vertical-layout theme="padding center">
         <h1>Increment, Decrement and Reset Button Example</h1>
         <td-counter></td-counter>
+        <nav>
```
flakolefluk, line 15: 👍🏾

starters/lit-mobx-vaadin/src/components/fetch/fetch.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    this.state.setFetchMessage('');
+  }
+
+  firstUpdated(
+    changedProperties: PropertyValueMap<any> | Map<PropertyKey, unknown>,
+  ) {
+    super.firstUpdated(changedProperties);
+
+    this.fetchMessage('lit-mobx-vaadin starter.dev!').then(message => {
```
flakolefluk, line 71: I generally prefer the page to be responsible for queryParams and requests and pass down props or it could be a better use case for the store. e.g the page makes the requests, and sets the message. another component reads from the store (in this case it's child component so it would be valid to use props too)

flakolefluk (changes requested):

starters/lit-mobx-vaadin/src/pages/page.mixin.ts
Amdrel: I made a primary color variable and also some link specific ones since they have unique colors separate from any palette. Let me know what you think :)

flakolefluk (approved):
🎉

PR #1262, opened 2023-05-15
[lit-mobx-vaadin] update deps, add test server, add unit tests

### LINKED ISSUE #1263: [lit-mobx-vaadin] add test runner + unit tests

flakolefluk:
## Type of change

* add test runner and unit tests
* updates deps to use latest

closes #1263 


starters/lit-mobx-vaadin/package.json
```diff
@@ -13,6 +13,7 @@
     "prettier": "eslint --ext .ts,.html . --fix --ignore-path .gitignore && prettier \"**/*.ts\" --write --ignore-path .gitignore",
     "storybook": "tsc && npm run analyze -- --exclude dist && concurrently -k -r \"tsc --watch --preserveWatchOutput\" \"wds -c .storybook/server.mjs\"",
     "storybook:build": "tsc && npm run analyze -- --exclude dist && build-storybook",
+    "test:unit": "tsc && wtr --coverage",
     "test": "playwright test",
```
Amdrel, line 17: I'm wondering if we should make the unit tests the default instead of the integration tests since as far as I know that's how all the other starts work.

Amdrel (changes requested):
There are some lint errors that need to be fixed before this can get merged. Also prettier needs to be run as well.

starters/lit-mobx-vaadin/package.json
flakolefluk: updated

Amdrel (approved):

Issue #1271, opened 2023-06-08
[Feature]: when initializing a kit, select the package manager


dustinsgoodman:
### Is your feature request related to a problem? Please describe.

When I init a project from starter.dev, I have to decide my package manager and generate the initial lock file.

### Describe the solution you'd like

When I run the CLI, I'd like to be prompted to select the package manager that is appropriate for my project.

For Node projects, I should be able to select npm, yarn, or pnpm.

For Deno or other projects, I should be presented with appropriate options for that kit.

### Describe alternatives you've considered

This may require a configuration file somewhere or extending an existing JSON file we have to enable this behavior.

PR #1280, opened 2023-06-12
feat: adding package manager option to CLI

jdwilkin4:
## Type of change

- [ ] Documentation change
- [ ] Bug fix
- [x] feature

## Checklist

- [ ] This fix resolves #

### DIFF
--- a/packages/create-starter/src/index.ts
+++ b/packages/create-starter/src/index.ts
@@ -10,6 +10,7 @@ import { trackSelectedKit } from './metrics';
 
 const STARTER_KITS_JSON_URL = 'https://raw.githubusercontent.com/thisdot/starter.dev/main/starter-kits.json';
 const EXCLUDED_PACKAGE_JSON_FIELDS = ['hasShowcase'];
+const PACKAGE_MANAGERS: string[] = ['npm', 'pnpm', 'yarn'];
 
 export async function main() {
   console.log(`\n${bold('Welcome to starter.dev!')} ${gray('(create-starter)')}`);
@@ -25,10 +26,12 @@ export async function main() {
     if (res.ok) {
       const starterKitsJSON = await res.json();
       if (typeof starterKitsJSON === 'object' && starterKitsJSON !== null) {
-        starters = Object.entries(starterKitsJSON).map(([name, description]) => ({
-          value: name as string,
-          title: description as string,
-        })).sort((a, b) => a.title.localeCompare(b.title));
+        starters = Object.entries(starterKitsJSON)
+          .map(([name, description]) => ({
+            value: name as string,
+            title: description as string,
+          }))
+          .sort((a, b) => a.title.localeCompare(b.title));
       }
     } else {
       throw new Error();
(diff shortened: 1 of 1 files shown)


packages/create-starter/src/index.ts
```diff
@@ -100,16 +107,22 @@ async function createStarter(options: prompts.Answers<'name' | 'kit'>): Promise<
       await initNodeProject(packageJsonPath, destPath, options);
     }
 
-    await initGitRepo(destPath);
+    await initGitRepo(destPath, packageManager);
     console.log(bold(green('✔') + ' Done!'));
     console.log('\nNext steps:');
     console.log(` ${bold(cyan(`cd ${options.name}`))}`);
```
jdwilkin4, line 113: I think Dustin wanted to update the language here to be this based on the slack thread disucssion:

oh update next steps to say "npm start" (or equivalent based on package manager selected" instead of "npm install"

I double check with him on that.

packages/create-starter/src/index.ts
```diff
@@ -44,35 +47,39 @@ export async function main() {
...
       type: 'text',
       name: 'name',
       message: 'What is the name of your project?',
     },
+    {
+      type: 'autocomplete',
+      name: 'packageManager',
+      message: 'Which package manager would you like to use?',
+      choices: PACKAGE_MANAGERS.map((pm) => ({ title: pm, value: pm })),
+      suggest: (input, choices) => Promise.resolve(choices.filter((c) => c.title.includes(input))),
```
jdwilkin4, line 62: How would this work for Deno?

I noticed this line in the [ticket](https://github.com/thisdot/starter.dev/issues/1271)

For Deno or other projects, I should be presented with appropriate options for that kit.

Is there a way to conditionally show package manger options based on the kit chosen?

packages/create-starter/src/index.ts
```diff
@@ -44,35 +47,39 @@ export async function main() {
       name: 'kit',
       message: 'Which starter kit would you like to use?',
       choices: starters,
-      suggest: (input, choices) => Promise.resolve(choices.filter(c => c.title.includes(input))),
+      suggest: (input, choices) => Promise.resolve(choices.filter((c) => c.title.includes(input))),
```
jdwilkin4, line 50: Once the merge conflict is resolved, then it will show the update for this filter function

packages/create-starter/src/utils.ts
```diff
@@ -15,12 +15,15 @@
   }
 }
 
-export async function initGitRepo(path: string) {
+export async function initGitRepo(path: string, packageManager: string) {
   return new Promise((resolve, reject) => {
-    exec(`git init ${path}`, (err) => {
+    exec(`cd ${path} && git init && ${packageManager} install && git add . && git commit -m 'init commit generated by starter.dev CLI' && cd ../`, (err, stdout) => {
```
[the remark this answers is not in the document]
jdwilkin4: I am having trouble on my end seeing the commit message
testing-angular-screenshot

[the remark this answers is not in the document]
jdwilkin4: I am looking closer at the `package.json` for the angular starter and it looks like there is no yarn dev command. That one has a yarn start command instead. 

That might pose some issues because some of the kits use `yarn dev` while others have `yarn start`. 

We could probably just shorten it to say `yarn` or `npm i`, etc to cover our bases

packages/create-starter/src/index.ts
WillHutt: When the kit is created we would still be outside of the newly created directory though and would still need to cd into it.

PR #1291, opened 2023-06-23
feat: updates card design for kits and homepage

WillHutt:
## Type of change

- [x] Feature
- [ ] Documentation change
- [ ] Bug fix

## Summary of change

- This updates the homepage and kit pages to reflect the option 1 selection in the Figma designs found in the Slack PTA channel
   - Updates the homepage to include more icons with a layout of 4x3 and adds a title for the kit
   - Adds a showcase section to individual kits with a View Showcase option that navigates to the section on the page
   - Adds updated keywords and icons from https://github.com/thisdot/starter.dev/issues/1283
   - Updates packages to reflect which kits have showcases
   - Resolves 2 failing checks

## Screenshots

Homepage
deploy-preview-1291--starter-dev netlify app_ (1)
deploy-preview-1291--starter-dev netlify app_

Individual kit
deploy-preview-1291--starter-dev netlify app_kits_next-react-query-tailwind_ (1)
deploy-preview-1291--starter-dev netlify app_kits_next-react-query-tailwind_

## Checklist

- [x] These changes follow the [contributing guidelines](https://github.com/thisdot/starter.dev/blob/main/CONTRIBUTING.md)
- [x] This starter kit has been approved by the maintainers
- [x] All dependencies are version locked
- [x] This fix resolves #1284 
- [x] I have verified the fix works and introduces no further errors

PR #1296, opened 2023-06-29
feat: adds auto generate showcases

### LINKED ISSUE #1292: [Pods] - Auto generate showcases
### Background
A concern was raised that if we don't auto generate it could easily get missed or become out of date. 

### Acceptance 
With that in mind the goal of this ticket is to adjust the scripts so that the package.json has a property to tell which showcases it has available and use that to help generate the repo links and such

- [x] Update scripts to grab showcase information if it is available

WillHutt:
## Type of change

- [x] Feature
- [ ] Documentation change
- [ ] Bug fix

## Summary of change

- Auto generates showcases based on package.jsons
- Refactored the original kit to generate script

## Checklist

- [ ] This fix resolves #1292 


hdJerry (dismissed):
Look good to me.
I Leave it for others to review

packages/website/scripts/generate.js
```diff
@@ -0,0 +1,137 @@
...
+# generated by scripts/kitgen at ${new Date().toISOString()}
+# edit KIT_PAGE_RELATIVE_LAYOUT_PATH in scripts/kitgen to change layout
+layout: ${KIT_PAGE_RELATIVE_LAYOUT_PATH}
+${frontmatter}
+---
+
+${markdown}`;
+}
+
+async function createDeleteFiles(kit, PagesPath) {
```
lindakatcodes, line 26: Since `kit` is a boolean value here, I'd suggest changing this to `isKit`, making it more obvious that it's not being passed in a kit but a true/false on if it is a kit or not.

packages/website/scripts/generate.js
```diff
@@ -0,0 +1,137 @@
...
+        kit
+          ? 'kitgen: creating kit pages directory'
+          : 'showcasegen: creating showcase file'
+      );
+      await fs.mkdir(PagesPath);
+    }
+  }
+}
+
+async function generate(kit, genPath) {
```
lindakatcodes, line 55: Same suggestion here: this feels better as a `isKit` name.

packages/website/src/showcases/showcases.json
```diff
@@ -0,0 +1,80 @@
+[
```
lindakatcodes, line 1: Since we're now completely generating this file, it should probably be added to the `.gitignore` file, like the kit pages are.

lindakatcodes (changes requested):
**This PR will likely be best held until the new design PR is approved and merged** - there's a reasonable bit of crossover between the two and having the showcases doesn't make sense without that new design. 

There will be a few places we'll need to adjust once the design PR (#1291 ) is merged:
- in `lib/parseKits`, I think we'll be able to adjust that to pick the showcases from the package.json instead of having the variable there
- the showcase variable will need to be removed from the config file
- in that showcase variable we're also setting the icon - that doesn't seem to be handled in the current PR, so will need to be solved again

I've left a few other questions and suggestions for this PR, most fairly minor. 

We should also update the root level README to indicate that new kits should include the "showcase" array, and what that should include / look like. 

I also wonder if we need to update the packages/website `predev` and `prebuild` scripts to also generate the showcases? We'll likely need that to ensure things are kept up to date as well.

WillHutt:
Yeah, I was initially planning on #1291 being merged before this was completed, but that had a few rounds of changes and is still ongoing. 

I was considering pushing this into #1291, but I also didn't want to hold that up with any feedback on this and cause a massive delay on it like in previous PRs/tickets. 

With that route chosen, I ended up removing the hook-up functionality I had originally implemented for instance
- The `lib/parseKits` adjustments
- Removal of the showcase variable from the config file
- The icon being handled since that isn't set in the package.json

I was going to create a new ticket that would focus on combining those above-needed changes with the new updates from #1291.

I'm good with either outcome and I will be pushing a new commit that should address a majority of the feedback you've given. 😄

lindakatcodes:
@WillHutt The design changes from #1291 is now merged into main! You should be good to rebase this PR when you come back and move forward from there. 👍🏼

lindakatcodes (commented):
Looks solid to me! The only tiny nitpick I'd make is - the array in each kit's package.json file, can we name it "showcases" instead of "showcase" (make it plural)? Since the idea is that we'll eventually have multiple showcases? Otherwise, this feels good!

lindakatcodes (approved):
:shipit:

